PR #20, opened 2018-01-17
vocabulary and pattern update

mjkoster:
1. remove "value" tag
2. add "const" tag"
3. fix remaining vocabulary issues readProperty, writeProperty
4. correct use of type and @type


mjkoster:
added tag is "const", not "cons"

mkovatsc:
(fixed typo directly in mjk's post instead of second comment)

@vcharpenay Could you check these changes in terms of the JSON Schema ontology? Maybe we do not require to distinguish between DataSchema and DataElement anymore (actual class names might be slightly different...)

vcharpenay:
I see two changes here: first, the `@type` statement is split into type parts and, second, you merge the `DataField` with `DataSchema` classes.

Let me address them separately on this example:
```js
"fields": [
    {
        "name": "level",
        "@type": ["sch:levelvalue"],
        "type": "integer",
        "min": 0,
        "max": 255
    },
```

_Note: what about relying on the deployed iotschema.org vocabulary for your examples? Here, `sch:levelvalue` corresponds to http://iotschema.org/CurrentLevel. There is a typo in the class definition (hence, the 404 page) but the class exists._

#### Type statement

`@type` is a JSON-LD reserved keyword that maps to [rdf:type](http://www.w3.org/1999/02/22-rdf-syntax-ns#). I assume that in the statement `_:subject rdf:type sch:levelvalue`, I assume that the statement subject (`_:subject`) is something like "203 with timestamp 2018-01-19T09:46:16+00:00", i.e. an actual level value.

The example schema definition above would map to RDF statements like:
```
_:subject
    rdf:type sch_levelvalue ;
    td:type td:Integer ;
    xsd:minInclusive "0" ;
    xsd:maxInclusive "255" .
```

in which, `_:subject` is the schema definition that validates "203 with timestamp 2018-01-19T09:46:16+00:00" and not the value itself. In fact, it would be more consistent if `@type` was `integer`, as defined [here](https://github.com/w3c/wot-thing-description/tree/master/ontology/td-types). For the same reason, `"@type": ["integer", "sch:levelvalue"]` would be an incorrect definition.

#### DataField vs. DataSchema

I am not sure that the change you propose here is consistent with the [JSON schema validation spec](http://json-schema.org/latest/json-schema-validation.html), where fields must have a name and a value, which in turn must be a valid JSON schema. The spec is hard to interpret, though. But anyway, to me, this change also limits the reuse of schema definitions. For instance, one could give a unique URI to the definition of a byte value:
```js
{
    "@id": "http://example.org/Byte",
    "name": "byte value",
    "@type": Integer,
    "minimum": 0,
    "maximum": 255
}
```

You would then simply declare the following in your TD, as defined the current spec:
```js
"fields": [
    {
        "name": "level",
        "value": "http://example.org/Byte"
    },
```

If you merge `DataField` and `DataSchema`, you cannot simply reference this new  `Byte` type.

PR #24, opened 2018-01-29
typo fix

yamagile:
typo fix


yamagile:
One more thing, I'm not sure what "faut domains" mean. Is "fault domain" correct?

mkovatsc:
Thanks! Merging.

PR #45, opened 2018-06-27
Mjkoster update bundang

mjkoster:
Update index.html with new TD format
Minor corrections


mkovatsc:
Merged as decided in main call 12 Sep 2018.

PR #51, opened 2019-04-14
Moving sequence diagrams from architecture spec to binding spec

mlagally:
See https://github.com/w3c/wot-architecture/issues/218
and corresponding PR.


egekorkan:
@mlagally I think the PR is mostly about adding the sequence charts but somehow the index.html is conflicting 100% even though I don't see how in the commit. Any thoughts?

PR #67, opened 2019-07-11
WIP: Updating to new TD spec and definitions

egekorkan:
These are the commits that correspond to updating the protocol binding to the CR version of the TD and Architecture specs

### DIFF
--- a/index.html
+++ b/index.html
@@ -559,14 +559,6 @@
             apply to the processing of the representations.
           </p>
           <p>
-            Additionally, there may be a profile which points to a URI for
-            further description, for example a form with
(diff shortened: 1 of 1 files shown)


egekorkan:
@mjkoster could you check if this PR is OK? Here is a rendered version: https://cdn.statically.io/gh/w3c/wot-binding-templates/korkan-update/index.html

mjkoster:
Yes, it looks good. 

Thanks!

Michael

mjkoster:
One more comment, in the CoAP and MQTT binding discussion, I would change "fictitious" to "unofficial" at some point.

egekorkan:
Regarding fictitious comment, I have copied it from TD spec so I would bring the discussion over ^^

Issue #81, opened 2019-12-02
Fixing CoAP vocabulary


egekorkan:
After discussing with Klaus Hartke from IETF @ektrah we need the following changes:
- Removing methodCode: This makes the TD less human readable and a client would need to understand the integer values anyways.
- Define features for CoAP (this thinking can be applied to other protocols): one example would be to define `cov:observe` in the CoAP vocabulary. This way the client doesn't have to understand what the header options are but simply use the feature (like a method in a CoAP library). If we look at the current stage from the point of view of an implementor who uses a CoAP library, he/she would need to understand what the option names and values mean and then remap them to the library. We will keep the optionName and optionNumber in case the server needs to describe custom headers but the idea is to not rely on these keywords to construct the request.
- For the future, he has has also pointed out that the keyword `subprotocol` is protocol specific anyways so why do we simply not define a subprotocol per protocol, like `"subprotocol:"cov:observe"` or `"subprotocol:"cov:multicast"`

egekorkan:
The TD spec has decided to do observing differently, i.e. use subprotocol like `"subprotocol" : "cov:observe"`.  This can be seen in [Example 33](https://w3c.github.io/wot-thing-description/#myLampThing-example-serialization)

egekorkan:
In the call of 13.12.2019, we have agreed to use the subprotocol field to describe observation and also putting a note not to use options and subprotocol at the same time.

egekorkan:
The changes are in the latest editor's draft.

PR #82, opened 2019-12-06
improve CoAP vocabulary to address #81

egekorkan:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-binding-templates/pull/82.html" title="Last updated on Dec 16, 2019, 6:41 PM UTC (e8517b1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-binding-templates/82/71e8851...e8517b1.html" title="Last updated on Dec 16, 2019, 6:41 PM UTC (e8517b1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -976,7 +976,7 @@
 
             "mqv:controlPacketValue": "SUBSCRIBE"
 
-            "cov:methodCode": 1
+            "cov:methodName": "GET"
           </pre>
 
         <p>
@@ -1484,6 +1484,11 @@
 
       <section>
         <h3>CoAP Vocabulary</h3>
+
+        <p>In the CoAP vocabulary, it is also possible to specify features of the protocol and detail the specific header options that need to be set.
+        This is indicated by <code>cov:feature</code> vocabulary term.
+        </p>
+
         <h4>CoAP Vocabulary Terms</h4>
         <table class="def">
           <thead>
@@ -1496,17 +1501,6 @@
           </thead>
           <tbody>
             <tr>
-              <td><code>cov:methodCode</code></td>
-              <td>CoAP method code (Literal).</td>
-              <td>optional</td>
-              <td>
-                <a
-                  href="https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#unsignedByte"><code>unsignedByte</code></a>
-                <p>(one of <code>1</code> (GET), <code>2</code> (POST), <code>3</code> (PUT), <code>4</code> (DELETE),
-                  <code>5</code> (FETCH), <code>6</code> (PATCH), <code>7</code> (iPATCH))</p>
-              </td>
-            </tr>
-            <tr>
               <td><code>cov:methodName</code></td>
               <td>CoAP method name (Literal).</td>
               <td>optional</td>
@@ -1517,29 +1511,27 @@
               </td>
             </tr>
             <tr>
+              <td><code>cov:feature</code></td>
+              <td>CoAP Feature (Literal), e.g. <code>"Observe"</code>
+              <td>optional</td>
+              <td>
+                <a href="http://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#string"><code>string</code></a>
+              </td>
+            </tr>
+            <tr>
               <td><code>cov:options</code></td>
-              <td>CoAP options sent with the message, e.g., <code>[ { "cov:optionNumber": 6, "cov:optionValue": 0 }]</code> to observe.</td>
+              <td>CoAP options sent with the message, e.g., <code>[ { "cov:optionName": "Observe", "cov:optionValue": 0 }]</code> to observe.</td>
               <td>optional</td>
               <td>
                 array of <code>cov:MessageOption</code>
               </td>
             </tr>
             <tr>
-              <td><code>cov:optionNumber</code></td>
-              <td>Option number (Literal), see <a
-                  href="https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#option-numbers">CoRE Parameters</a>.</td>
-              <td>mandatory within <code>cov:MessageOption</code> if <code>cov:optionName</code> is not present</td>
-              <td>
-                <a
-                  href="https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#unsignedShort"><code>unsignedShort</code></a>
-              </td>
-            </tr>
-            <tr>
               <td><code>cov:optionName</code></td>
               <td>Option name (Literal), see <a
-                  href="https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#option-numbers">CoRE
+                  href="https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#option-names">CoRE
                   Parameters</a>.</td>
-              <td>mandatory within <code>cov:MessageOption</code> if <code>cov:optionNumber</code> is not present</td>
+              <td>mandatory within <code>cov:MessageOption</code></td>
               <td>
                 <a href="http://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#string"><code>string</code></a>
               </td>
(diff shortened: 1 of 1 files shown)


egekorkan:
@ektrah can you also have a look?

egekorkan:
This is now the final version before going to publication process. Please review

index.html
```diff
@@ -1349,7 +1349,7 @@ <h3>Event Forms</h3>
       </section>
     </section>
 
-    <section>
+    section>
```
ektrah:
```suggestion
    <section>
```

index.html
```diff
@@ -1518,28 +1512,18 @@ <h4>CoAP Vocabulary Terms</h4>
             </tr>
             <tr>
               <td><code>cov:options</code></td>
-              <td>CoAP options sent with the message, e.g., <code>[ { "cov:optionNumber": 6, "cov:optionValue": 0 }]</code> to observe.</td>
+              <td>CoAP options sent with the message, e.g., <code>[ { "cov:optionName": "Observe", "cov:optionValue": 0 }]</code> to observe.</td>
```
ektrah: Since Observe is more suitable to be done with `"subprotocol":"cov:observe"`, should this example better use a different CoAP option?

index.html
```diff
@@ -1566,61 +1551,70 @@ <h4>CoAP Default Vocabulary Terms</h4>
...
             </tr>
             <tr>
               <td><code>writeproperty</code></td>
-              <td><code>"cov:methodCode": 3</code> OR <code>"cov:methodName": "PUT"</code></td>
+              <td><code>"cov:methodName": "PUT"</code></td>
             </tr>
             <tr>
               <td><code>observeproperty</code></td>
-              <td> <code>"cov:methodCode": 1</code> OR <code>"cov:methodName": "GET"</code> with <code>[ { "cov:optionName": "Observe", "cov:optionValue": 0 }]</code></td>
+              <td><code>"subprotocol":"cov:observe"</code></td>
```
ektrah: Since Observe can also be used with the FETCH method, it might be good to specify `"cov:methodName": "GET"` in addition to `"subprotocol":"cov:observe"`.

index.html
```diff
@@ -1566,61 +1551,70 @@ <h4>CoAP Default Vocabulary Terms</h4>
...
         </table>
+        </section>
+
+        Observing Resources in CoAP should be done as specified in [[!RFC7641]].
+        Since observing and unobserving need to be indicated with the Observe flag in the header, this mechanism can be also described in the <code> cov:options </code> with the value <code>[ { "cov:optionName": "Observe", "cov:optionValue": 0 }]</code>. 
+        However, this is not enough to describe the mechanism of getting asynchronous updates from the observed resource.
+        Thus, <code>"subprotocol":"cov:observe"</code> is used to indicate the observation mechanism.
+
+        <p class="ednote">
+          As indicated by the <a href="#coap-default-vocabulary-terms" class="sec-ref">CoAP Default Vocabulary Terms</a>, it is recommended to use the <code>subprotocol</code> to describe the observation and using both the <code> cov:options </code> and <code>subprotocol</code> can lead to confusion.
```
ektrah: For that reason, I would probably say `MUST NOT` on `[ { "cov:optionName": "Observe", "cov:optionValue": 0 }]`.

index.html
```diff
@@ -1740,14 +1738,15 @@ <h3><code>subprotocol</code> Vocabulary</h3>
...
-          for HTTP.
+          Currently, the supported values are <code>longpoll</code>, <code>websub</code> and <code>sse</code> defined for HTTP.
           Subprotocols can be used for asynchronous event delivery or observing Properties.
         </p>
         <p>
           For WebSockets, the IANA-registered Websocket Subprotocols [[iana-web-socket-registry]] may be used.
         </p>
-
+        <p>
+          For CoAP, <code>"subprotocol":"cov:observe"</code> can be used to describe asynchronous observation operations as defined by [[RFC6741]]
```
ektrah: Just curious: I assume that `cov:observe` gets expanded to `http://www.example.org/coap-binding#observe`. What do `longpoll`, `websub`, and `sse` get expanded to?

vcharpenay: At the moment, they are defined as simple strings, see https://www.w3.org/2019/wot/hypermedia#forsubprotocol. Ideally, the HTTP vocabulary in RDF should define URIs for these sub-protocols.

vcharpenay: One issue with JSON-LD expansion (and transformation to RDF) is that we cannot have both strings and URIs in expanded form for a single context. Currently, in TDs, `cov:observes` will expand to the string literal `"cov:observe"` and not the expected URI.

ektrah: I see. I think implementers might get confused why all the other `cov:`'s get expanded but this one doesn't. Maybe `coap-observe` or something similar would be preferable to `cov:observe`?

index.html
egekorkan: I agree :) What options would be useful and give-able in such a small example? I have thought of  `If-Match` or `Max-Age`

index.html
egekorkan: Done with https://github.com/w3c/wot-binding-templates/pull/82/commits/8d36578677e1a4fac6b81c198bae6615cb02b4af

index.html
ektrah: `[ { "cov:optionName": "Accept", "cov:optionValue": 110 }]` maybe? (Max-Age is not valid in requests. Don’t know if If-Match is a good or bad example.)

One interesting question here is: If the consumer does not understand what `”Accept”` means and is just supposed to include this option with this name in its request (which is legitimate in case of `Accept`), how does it find out that the option number is 17 and the option value format is `uint`?

index.html
egekorkan: Addressed in https://github.com/w3c/wot-binding-templates/pull/82/commits/3ff9303538e65e12ef1b3e48f0e7961940d023af

egekorkan: However, I am not sure if this is what you meant

index.html
egekorkan: I find your proposal better ^^ and put it in https://github.com/w3c/wot-binding-templates/pull/82/commits/e8517b122888c30bffe30b1b7324539cd0d049f5

egekorkan: Regarding the question, I would suggest that the client should have a mapping of these names to the actual protocol header values. I have looked at [node-coap](https://github.com/mcollina/node-coap#setOption) implementation where one would actually use the names in the API and the value also as a string instead of the `uint`. I think that similar to the `methodName`, the implementations just use the more human-readable way to specify things.

index.html
egekorkan: In the case `coap-observe`, this should be then defined in the TD vocabulary?

index.html
ektrah: Right now (even with the update) it sounds like a consumer can freely choose between the subprotocol and `[ { "cov:optionName": "Observe", "cov:optionValue": 0 }]` (with the hard requirement not to have both and a soft preference for the former). IMO we should forbid the latter entirely, as I cannot imagine any scenario where it would be useful that a consumer sets the Observe option without understanding what it's doing. (That's how we defined `cov:options`, right?)

index.html
ektrah: But wasn't the scenario for `cov:options` that the client is unaware of that CoAP option? If the client is aware, why not simply `"cov:ocf-content-format": 12345` instead of `"cov:options": [ { "cov:optionName": "OCF-Content-Format", "cov:optionValue": 12345 }]`? For the method, the client always has to be aware of what it means, so that's different.

egekorkan: I think this is a question of whether the client is aware that `ocf-content-format` is an option or not. I am not sure why this array notation was picked in the beginning. Probably to give the hint that "these are options, go figure out how to put them into your request". 

I think that having them inside the array looks more contained but bulkier. In the end, it doesn't help the fact that the client has to know how the options map to its client implementation. I still think that a more human readable way is better since some libraries that build the requests with a function that gets the options in the `optionName` kind of way. Golang implementation and the node.js implementation do this for example. Whereas some libraries will do the `optionNumber` kind of way. Californium is one example, see [here](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java#L498). So I think it would be better to have `name` instead of `number` since we can somewhat assume that a user of a library with `number` would have more knowledge of the protocols.

egekorkan:
I will be merging this PR to simplify the reviewing process, new issues can be thus opened against it

PR #98, opened 2020-07-09
Modbus Vocabulary

Drukob:
As a part of my Master's Thesis at TUM, I created a Modbus RDF Vocabulary that allows you to describe (legacy) systems with a Modbus interface, such as industrial PLCs, with TD. The thesis was supervised by +@egekorkan and partially written with a software company in Garching. I added a simplified example of a Modbus Thing which was developed during the thesis and included it in the index.html, as well as a reference to the official Modbus specifcation. The modbus.ttl file was included in render.js to generate the corresponding html file for the vocabulary.

The Modbus protocol is not a typical IoT protocol, however still widely used in industrial communication, due to its simplicity and easy implementation. As it was never designed or intended as an official standard, implementations can vary in certain areas, therefore a specification of Modbus messages in RDF/ttl makes a lot of sense. For TD, further support of such industrial fieldbus protocols (Modbus, Profinet, Ethercat, etc.) would increase its applicability and acceptance in the industry.

***
<a href="https://pr-preview.s3.amazonaws.com/Drukob/wot-binding-templates/pull/98.html" title="Last updated on Mar 5, 2021, 1:58 PM UTC (e4bbf48)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-binding-templates/98/d1a7c53...Drukob:e4bbf48.html" title="Last updated on Mar 5, 2021, 1:58 PM UTC (e4bbf48)">Diff</a>


egekorkan:
Thank you @Drukob ! In addition to my comments above, it is also good to have a table in the index.html file that lists the vocabulary options, or at least link to the vocabulary. What is definitely needed is to propose a default binding for WoT operations (`readproperty`, `invokeaction` etc.) to Modbus operations (`"modv:FunctionCodes": "FC2"` etc.)

sebastiankb:
there were some comments in today's TD call: 

- the Modbus example is not updated in the index.html file
- we should evaluate if the modbus term a common used in other modbus implementation (e.g., length vs. byteLength vs number of register, offset address etc)
- we should also evaluate this approach with the node-wot modbus binding: https://github.com/eclipse/thingweb.node-wot/tree/master/packages/binding-modbus
- we should evaluate if the byte length is provided in the content type

sebastiankb:
Some more comment or question. The [Modbus implementation](https://github.com/eclipse/thingweb.node-wot/tree/master/packages/binding-modbus) comes with a `modbus:timeout`. It seems, this is missing here. 

The question should be also answered, what this timeout mean like "Client is not allowed to do a modbus read until timeout" vs "client has time until the timeout to do a read".

Maybe @relu91 can answer that?

relu91:
> "client has time until the timeout to do a read"

In the implementation, `modbus:timeout` refers to the total amount of time that the Modbus client should wait for an answer. So it is similar to the quoted statement. 

 On the other hand, this feature:
 
> "Client is not allowed to do a modbus read until timeout"

It is kind of covered by [modbus:pollingTime](https://github.com/eclipse/thingweb.node-wot/tree/master/packages/binding-modbus#modbuspollingtime).

sebastiankb:
As decided in today's TD call:
- we would like to provide side binding documents for the concrete protocols 
- modbus can be the first prototype where we can see how this can look like
- @mjkoster support this idea
- @wiresio may also support this 
- @relu91 may be also interested in this

relu91:
I also support the idea! +1 from my side. I guess there are still some misalignments with the implementation(the related issue in node-wot linked above), but we could solve them while we produce the new document.

PR #99, opened 2020-07-22
Presentation ROS as addition to the W3C WoT

MarcusWolfSchmidt:


egekorkan:
Since the presentation was already during a call of the task force, I am merging it now.

PR #109, opened 2021-02-03
Refining Modbus protocol binding

relu91:
This PR is based on the work of #98 and it focuses on the vocabulary alignment with the current [node-wot modbus implementation](https://github.com/eclipse/thingweb.node-wot/tree/3e9ccd44072a097d0452b5c306819c6d8fb04e12/packages/binding-modbus) (see also https://github.com/eclipse/thingweb.node-wot/issues/297). It fixes also the main example provided in the index.html. 

Known open issues:
- The implementation uses [range](https://github.com/eclipse/thingweb.node-wot/tree/3e9ccd44072a097d0452b5c306819c6d8fb04e12/packages/binding-modbus#modbusrange) to specify the starting and the end of the reading/writing operation. The vocabulary uses `offset` and `length`
- This proposal does not follow the new guidelines for a protocol binding description

### DIFF
--- a/ontology/modbus.ttl
+++ b/ontology/modbus.ttl
@@ -65,7 +65,16 @@ _:jakob a schema:Person ;
 				rdfs:label "Function Code"@en ;
 
 				rdfs:subClassOf :Code, schema:Enumeration .
-	
+
+	:FunctionName rdf:type owl:Class ;
+				rdfs:comment "A Modbus human readable function name"@en ;
+				rdfs:label "Function Name"@en ;
+				rdfs:subClassOf schema:Enumeration .
+
+	:Entity rdf:type owl:Class ;
+				rdfs:comment "A Modbus registry type"@en ;
+				rdfs:label "Entity"@en ;
+				rdfs:subClassOf schema:Enumeration .
 	
 	:ResponseCode rdf:type owl:Class ;
 				  rdfs:comment "Response Code"@en ;
@@ -78,12 +87,19 @@ _:jakob a schema:Person ;
 	
 	# ###  http://www.example.com/ns/modbus#options
 	
-	:functionCode   rdf:type owl:ObjectProperty ;
+	:function   rdf:type owl:ObjectProperty ;
 					rdfs:isDefinedBy <> ;
 					rdfs:comment "Function Code sent by the master in every request. Specifying the desired interaction."@en ;
-					rdfs:label "Function Code"@en ;
+					rdfs:label "Function"@en ;
 					rdfs:domain :Request ;
-					rdfs:range :FunctionCode .
+					rdfs:range :FunctionCode, :FunctionName .
+
+	:entity   rdf:type owl:ObjectProperty ;
+					rdfs:isDefinedBy <> ;
+					rdfs:comment "A registry type to let the runtime automatically detect the right function code"@en ;
+					rdfs:label "Entity"@en ;
+					rdfs:domain :Request ;
+					rdfs:range :Entity .
 	
 	
 	:responseCode rdf:type owl:ObjectProperty ;
@@ -151,6 +167,16 @@ _:jakob a schema:Person ;
 						 rdf:range xsd:integer ;
 						 rdfs:domain :RequestOption;
 						 om:usesUnit om:milliseconds.
+
+	:functionName		 rdf:type owl:ObjectProperty;
+						 rdfs:label "Function name of a function Code"@en;
+						 rdf:range :FunctionName ;
+						 rdfs:domain :FunctionCode.
+
+	:functionCode		 rdf:type owl:ObjectProperty;
+						 rdfs:label "Function code of a function Name"@en;
+						 rdf:range :FunctionCode ;
+						 rdfs:domain :FunctionName.
 	
 	#################################################################
 	#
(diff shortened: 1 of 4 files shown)


sebastiankb:
@vcharpenay can you support @relu91 to run the render script?

sebastiankb:
From today's TD call:
- the provided ontology file should be "TD" independent
- we would need a separate modbus-wot binding document which provides the mapping from TD operations like readproperty to Modbus operation 
- this document should be also have couple of examples that shows how TD should look like with modbus binding
- @egekorkan can you support here for the structure of the document?

mjkoster:
Does the node-wot modbus driver require "contentType": "application/octet-stream" to be present in the form?

relu91:
> Does the node-wot modbus driver require "contentType": "application/octet-stream" to be present in the form?

Currently, it basically ignores any cotentType provided in the form and it overrides it with `application/octet-stream`. Not sure if TD specs require to fail if a binding cannot fulfill the contentType requested. HTTP treats it as a suggestion, right?

mjkoster:
The documentation at modbus dot org consistently uses "address" (starting address, output address,...) in the protocol description. Is there a compelling reason to use "offset" in this vocabulary? Sorry (if I missed the discussion just point me there), but there doesn't seem to be a specific issue in the issue tracker...

mjkoster:
In creating the forms for a TD, I found it redundant to specify both offset (address) and include it in the URI scheme. 

I may also have missed the discussion of URI scheme, but I think we should consider just the part specified by modbus e.g.:
modbus+tcp://192.168.1.1:502/
since the modbus tcp ADU includes the unitID, function code and address.

If we want to extend the URI scheme, it seems like we should consider including unitID and function code, since they are part of the addressing model in modbus:

modbus+tcp://192.168.1.1:502/?unitid=1&function=3&address=1000&quantity=7

mjkoster:
For the multiple register/coil/input operations, we could consider using "quantity" to indicate how many 16 bit items are included in the payload. The term "length" is used consistently in the documents at modbus dot org to indicate total (ADU?) size in bytes

relu91:
Great points @mjkoster let me answer below:
>The documentation at modbus dot org consistently uses "address" (starting address, output address,...) in the protocol description. Is there a compelling reason to use "offset" in this vocabulary? Sorry (if I missed the discussion just point me there), but there doesn't seem to be a specific issue in the issue tracker...

Actually, this was @ga82qok's choice.  As you're suggesting it is better to move our notation to common concepts in the Modbus dot org. I could address this change in this PR or in subsequent PRs.

> In creating the forms for a TD, I found it redundant to specify both offset (address) and include it in the URI scheme.
>I may also have missed the discussion of URI scheme, but I think we should consider just the part specified by modbus e.g.:
modbus+tcp://192.168.1.1:502/
since the modbus tcp ADU includes the unitID, function code and address.
>If we want to extend the URI scheme, it seems like we should consider including unitID and function code, since they are part of the addressing model in modbus:
>modbus+tcp://192.168.1.1:502/?unitid=1&function=3&address=1000&quantity=7
>The use of URI parameters would IMO be a slightly better practice than assigning slash-separated segments in a well-known order

In [node-wot implementation](https://github.com/eclipse/thingweb.node-wot/tree/master/packages/binding-modbus) we are using the following syntax for modbus URLs:
```
modbus+tcp:// <host> [ : <port> ] [/ <unitid> [ ?offset=<offset> [&length=<length> ] ] ]
```
Putting aside the naming issue of offset and range, the rationale was to follow a rest-like structure. Function codes are not included because I think they're more verbs (e.g., like GET for HTTP) than addresses but I may misinterpret the Modbus philosophy there. 

> For the multiple register/coil/input operations, we could consider using "quantity" to indicate how many 16 bit items are included in the payload. The term "length" is used consistently in the documents at modbus dot org to indicate total (ADU size -6 bytes?) size in bytes, and they use "quantity" (of Registers, of Outputs...) to refer to the field in the protocol spec and the PDU example in the TCP spec. Is there a compelling reason to use "length" in this vocabulary to refer to uses of the term "quantity" in the modbus spec?

No particular reason thanks again for reviewing the Modbus documentation and reporting the correct term.

relu91:
Current status, from @sebastiankb  points:
- [X] the provided ontology file should be "TD" independent
- [X] we would need a separate modbus-wot binding document which provides the mapping from TD operations like readproperty to Modbus operation
- [ ] this document should be also have couple of examples that shows how TD should look like with modbus binding

The last TODO is almost there. I moved the example from the main document but we should provide more specific examples to ease the adoption from newbie developers. 

Other TODOs:
- [ ] Cover the empty sections in the document (e.g., introduction, entity, abstract, etc.)
- [ ] Correctly link to Modbus ontology
- [ ] Improve mappings table
- [ ] Improve ontology mapping description

relu91:
I improved a little bit this PR by adding new examples and descriptions about `Entity`, `FunctionName`, and `FunctionCode`. There is still room for improvements but it is a good first draft.

bindings/modbus/index.template.html
```diff
@@ -0,0 +1,338 @@
...
+                  "base": "",
+                  "properties": {
+                      "limitSwitch1": {
+                          "title": "downLimitSwitch",
+                          "type": "boolean",
+                          "description": "Limit switch moving downwards",
+                          "forms": [
+                              {
+                                  "op": "readproperty",
+                                  "href": "modbus+tcp://192.168.178.32:502/10003",
```
sebastiankb, line 265: URL should be 192.168.178.32:502/1

bindings/modbus/index.template.html
```diff
@@ -0,0 +1,338 @@
...
+                          ]
+                      },
+                      "limitSwitch2": {
+                          "title": "forwardLimitSwitch",
+                          "type": "boolean",
+                          "description": "Limit Switch moving forward",
+                          "forms": [
+                              {
+                                  "op": "readproperty",
+                                  "href": "modbus+tcp://192.168.178.32:502/10002",
```
sebastiankb, line 281: same here, offset is used instead unitID

bindings/modbus/index.template.html
```diff
@@ -0,0 +1,338 @@
...
+                          "type": "boolean",
+                          "description": "Down Motor Status (single coil). PLC output, can be written to control the motor.",
+                          "forms": [
+                              {
+                                  "op": [
+                                      "writeproperty",
+                                      "observeproperty",
+                                      "readproperty"
+                                  ],
+                                  "href": "modbus+tcp://192.168.178.32:502/00006",
```
sebastiankb, line 301: again

bindings/modbus/index.template.html
```diff
@@ -0,0 +1,338 @@
...
+                          "type": "boolean",
+                          "description": "Forward Motor Status (single coil). PLC output, can be written to control the motor.",
+                          "forms": [
+                              {
+                                   "op": [
+                                      "writeproperty",
+                                      "observeproperty",
+                                      "readproperty"
+                                  ],
+                                  "href": "modbus+tcp://192.168.178.32:502/00003",
```
sebastiankb, line 322: again

sebastiankb (changes requested):

ontology/modbus.ttl
```diff
@@ -0,0 +1,358 @@
...
+				  rdfs:label "Response Code"@en ;
+
+				  rdfs:subClassOf :Code, schema:Enumeration  .
+	
+	
+	### Properties
+	
+	# ###  http://www.example.com/ns/modbus#options
+	
+	:function   rdf:type owl:ObjectProperty ;
```
sebastiankb, line 91: I would recommend to introduce functionCode and functionName. One of them can be used. Alternatively, both can be used at the same time, but should not contradict each other.

bindings/modbus/index.template.html
```diff
@@ -0,0 +1,338 @@
...
+                          "scheme": "nosec"
+                      }
+                  },
+                  "security": "nosec_sc",
+              
+                  "base": "",
+                  "properties": {
+                      "limitSwitch1": {
+                          "title": "downLimitSwitch",
+                          "type": "boolean",
```
sebastiankb, line 260: beside the Data schema types we should always include @type and with a precise type annotation like unsignedLong, double, etc. We can rely on XSD

bindings/modbus/index.template.html
```diff
@@ -0,0 +1,338 @@
...
+                          "description": "Limit switch moving downwards",
+                          "forms": [
+                              {
+                                  "op": "readproperty",
+                                  "href": "modbus+tcp://192.168.178.32:502/10003",
+                                  "modv:function": "2",
+                                  "modv:length": 1,
+                                  "modv:offset": "10003",
+                                  "modv:unitID": "1",
+                                  "contentType": "application/octet-stream"
```
sebastiankb, line 270: contentType should also include the order pattern of the bytes in the registers. E.g., lets assume the we receive the bytes in this order: 1 | 2 | 3 | 4 
Decoding a value (e.g., float) with little endian would lead to a reading sequence: 4 3 2 1
Sometimes the endians are only register wise: In that case we whould have an little endian order like 2 1 4 3 

I would suggest to introduce a parameter byteSeq which provides the encoded order of the registers. Eg.:

`"contentType": "application/octet-stream;byteSeq=1234"`

bindings/modbus/index.template.html
sebastiankb: from today's TD call: @type is maybe not the best solution. We should evaluate what JSON Schema provides and/or should introduce another term like "subtype".

bindings/modbus/mapping.ttl
```diff
@@ -0,0 +1,56 @@
...
+@prefix om: <http://www.ontology-of-units-of-measure.org/resource/om-2/> .
+@prefix dct: <http://purl.org/dc/terms/> .
+@prefix owl: <http://www.w3.org/2002/07/owl#> .
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix xml: <http://www.w3.org/XML/1998/namespace> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix vann: <http://purl.org/vocab/vann/> .
+@prefix schema: <http://schema.org/> .
+@prefix binding: <https://www.w3.org/2019/wot/binding#> .
```
sebastiankb, line 11: from today's TD call: we should clearify if this is needed. Maybe the SHACL file from the TD can be also be used.

OhArthits (approved):

bindings/modbus/mapping.ttl
relu91: Now we are using SHACL.

egekorkan:
In the call of 14.07.2021, we have decided to merge this and create PRs to fix some small discussions.

Issue #139, opened 2021-12-03
XML template is more for green field devices


relu91:
Reviewing [XML Binding template](https://w3c.github.io/wot-binding-templates/bindings/payloads/xml/index.html) document I noticed that it is trying more to describe a way to serialize JSON data to XML rather than trying to describe existing XML documents with our data schema. Given that I can't really find any modern API that is returning XML data I think we should be more open and don't force to use a particular "serialization" pattern even for XML. 

Taking for example this XML payload:
```xml
<SaveContactResponse xmlns="http://schemas.datacontract.org/2004/07/SmashFly.WebServices.ContactManagerService.v2"› 
    <ContactId>2147483647</ContactId>
    <Errors>
        <string xmlns="http://schemas.microsoft.com/2003/10/Serialization/Arrays">Error 1</string>
        <string xmlns="http://schemas.microsoft.com/2003/10/Serialization/Arrays">Error 2</string>
    </Errors>
    <HasErrors>true</HasErrors>
</SaveContactResponse>
```
I think we can correctly map to this JSON object:
```js
{
  SaveContactResponse: {
    ContectId : 2147483647,
    Errors: [ { string: "Error 1" }, { string: "Error 2"} ],
    HasErrors: true,
  }
}
```
Consequently, knowing this mapping, we can define the correct data schema:
```js
{
  type: "object"
  properties: {
    "SaveContactResponse": { type: "object" /* etc. */}
  }
}
```

I think that giving a clear algorithm to map an XML document to a JSON document is better than the current specification, maybe we can even find something already defined. Opinions?

danielpeintner:
> I think that giving a clear algorithm to map an XML document to a JSON document is better than the current specification, maybe we can even find something already defined. Opinions?

The current algorithm is essentially inspired? from [EXI for JSON](https://www.w3.org/XML/EXI/docs/json/exi-for-json.html).

It maps JSON to XML.

What you are describing is mapping XML to JSON.
I think this is not possible in a *generic* way since XML is more powerful, e.g., XML has the power to
* have attributes and character for a simple type  element `<ContactId hasFoo="true">2147483647</ContactId>`
* well-defined sequences of elements ...
* use `xsi:type` casting
* use `nillable`  in an instance to indicate that there is *no* content..
* ...

Hence, I don't think this will ever work consistently.. 

w.r.t. *describe* XML to JSON conversion I am not sure how to move on from here best.

egekorkan:
[...]

PR #140, opened 2021-12-03
Introduce draft for text binding template

relu91:
I've started working on new documents for different payload binding templates. As suggested in #139 it might be easier to have a section that explains how to map data (conformant to a particular content type) to "Data Schema values". Informally we can think of "Data Schema values" as JavaScript values, but I would like to have them rigorously defined somewhere. 

This PR is of course still heavily WIP, but I'd like to have feedback as soon as possible about the new document structure.

### DIFF
--- a/bindings/payloads/text/index.html
+++ b/bindings/payloads/text/index.html
@@ -1,0 +1,130 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <meta charset='utf-8'>
+    <script src='https://www.w3.org/Tools/respec/respec-w3c' class='remove'></script>
+    <script class='remove'>
+        var respecConfig = {
+            specStatus: "ED",
+            editors: [
+                {
+                    name: "Cristiano Aguzzi",
+                    w3cid: "105495",
+                    company: "Invited Expert",
+                    companyURL: "https://github.com/relu91"
+                }
+            ],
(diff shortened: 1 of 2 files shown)


relu91:
I'm still grinding on this. I've updated a little bit the structure and text (also in the main document). TODOs:
- [] correctly cite our own terms (Thing Description, Data Schema etc.)
- [] Discuss the changes in Payload Representation section

egekorkan:
I think that it looks good overall. For changes to the main document, I think it is needed to refer to https://w3c.github.io/wot-thing-description/#sec-data-schema-vocabulary-definition
For changes to the text binding, the most "frustrating" part is that the word text is used as a non-technical word as well (a bit like Thing). So when you say text binding or similar, we should agree on what to use. Some suggestions could be always saying:
1. Text Plain (done once)
2. `text/plain`
3. Text with capital T and putting a note somewhere saying something like "The word Text when used with a capital T means the `text/plain` media type"

I see some typo or english mistakes but I will go over when the document is more stable. Also, Fujitsu had a Thing with this media type. One example I could find is here: https://github.com/w3c/wot-testing/blob/main/events/plugfest/2019-tpac-fukuoka/TDs/Fujitsu/local/Fujitsu-AgriDevice.jsonld#L61

@mryuichi could you share your experience with using this media type in TDs?

Issue #49, opened 2018-10-03
CoAP Blockwise Indicator


mkovatsc:
It would be good to have metadata in the TD that tells from which payload size blockwise transfers should be used.

In addition -- or combined, there should be metadata if blockwise transfers are supported at all by the Thing.

egekorkan:
@mkovatsc could you provide an example on how this should look like? We said that this metadata format should be provided in the CoAP vocabulary and the binding templates references to it and gives an example.

takuki:
On September 20 in Fukuoka, it was suggested to discuss this with IETF during Singapore F2F meeting. See minutes [here](https://www.w3.org/2019/09/18-wot-minutes.html#item13).

JKRhb:
Adding on to this issue (I noticed it because of https://github.com/w3c/wot-discovery/issues/93), I think you could theoretically add a CoAP Block option like the following to a Form to indicate that a blocksize of 64 is going to be used in responses:

```json
{
  "cov:optionName": "Block2",
  "cov:optionValue": {
    "num": 0,
    "m": 0,
    "szx": 64
  }
}
```

Maybe this could also be abbreviated using something like `cov:block2Size` or `cov:block1Size`. As the actual blocksize is negotiated between client and server during the message exchange, however, I am not sure if this is actually necessary.

egekorkan:
> As the actual blocksize is negotiated between client and server during the message exchange, however, I am not sure if this is actually necessary.

If this is indeed always the case, I do not think it makes sense. Can the client use this information in any way, like starting the negotiation with a better initial value?

JKRhb:
> Can the client use this information in any way, like starting the negotiation with a better initial value?

Oh, I think this might actually be a valid use case (c. f.  [figure 4 in RFC 7959](https://datatracker.ietf.org/doc/html/rfc7959#section-3.1)). This way it could be avoided that a consumer is surprised by a Block2 size that is too large. Thinking about it, this could be especially relevant for Block1 (POST/PUT) requests where the Thing might prefer a smaller block size (c. f. [figure 9](https://datatracker.ietf.org/doc/html/rfc7959#section-3.2)), e. g. because it is not able to process larger packets.

I think the main question here would then be how to deal with the Block option format where three values are encoded into one unsigned integer. To follow the current format, the example I gave above should probably actually look like this:

```json
{
  "cov:optionName": "Block2",
  "cov:optionValue": 2
}
```

[...]

Issue #143, opened 2022-01-06
"Protocol Binding" vs. "Binding Template"


benfrancis:
I've noticed that the current draft of the WoT Binding Templates specification uses the terms "protocol binding" and "binding template" fairly interchangeably. In actual fact they are two distinct concepts, with one making use of the other.

From the WoT Binding Templates specification
> Binding Templates enable a Thing Description to be adapted to a specific protocol, data payload formats or platforms that combine both in specific ways. This is done through additional descriptive vocabularies, Thing Models and examples that aim to guide the implementors of Things and Consumers alike. This document acts as the base and explains how other binding templates should be designed. Concrete binding templates are then provided in their respective documents that are linked to from this document.

From the WoT Thing Description specification:
> A Protocol Binding is the mapping from an Interaction Affordance to concrete messages of a specific protocol... Protocol Bindings of Interaction Affordances are serialized as forms

My understanding is that basically a binding template provides a vocabulary for describing a protocol binding with a given protocol, and a protocol binding is an instance of that template for an individual thing.

With the creation of the WoT Profile specification there are now two types of protocol binding:
1. Declarative - Serialised as forms in a Thing Description instance, using vocabulary from a binding template (e.g. the [HTTP Binding Template](https://w3c.github.io/wot-binding-templates/bindings/protocols/http/index.html))
2. Concrete - Referenced from a standardised protocol binding provided by a profile specification (e.g. the [Core Profile](https://w3c.github.io/wot-profile/#core-profile)), denoted by the profile member of a Thing Description

This makes some of the WoT Binding Templates specification a bit confusing to read. For example the title of [section 4.1.1](https://w3c.github.io/wot-binding-templates/#creating-a-new-protocol-binding) is "Creating a new Protocol Binding" whereas it should probably be called "Creating a new Binding Template".

I suggest reviewing the wording in the Binding Templates specification to make sure the distinction between a "protocol binding" and a "binding template" is clear.

egekorkan:
I think I am partly to blame for the confusion in general since I made it apparent that there are these two terms. The way the new structure is aiming to handle this is:
- The reason behind the word `template` is not known to me. My guess is that these binding specifications are recommendations on how to use certain protocols/platforms. What you are saying at

> My understanding is that basically a binding template provides a vocabulary for describing a protocol binding with a given protocol, and a protocol binding is an instance of that template for an individual thing.

[...]

PR #149, opened 2022-02-24
CoAP Binding based on features

ektrah:
First draft, based on the discussion in #49 

/cc @JKRhb @egekorkan @chrysn @mkovatsc

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Feb 24, 2022, 2:41 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fwot-binding-templates%2Fd95d21c3b57ba9f83416036c87ffb16b0e6a20e6%2Findex.html%3FisPreview%3Dtrue)

```
🕵️‍♀️  That doesn't seem to be a ReSpec document. Please check manually: https://labs.w3.org/spec-generator/uploads/6V7fna/tmp/?isPreview=true&publishDate=2022-02-24
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/wot-binding-templates%23149.)._
</details>

### DIFF
--- a/bindings/protocols/coap/index.html
+++ b/bindings/protocols/coap/index.html
@@ -95,7 +95,7 @@
                 </tr>
                 <tr>
                     <td>Quick Block-Wise Transfer</td>
-                    <td><code>"coap:quickblockwise": { ... }</code></td>
+                    <td><code>"coap:qblockwise": { ... }</code></td>
                 </tr>
                 <tr>
                     <td>Hop Limit</td>
@@ -117,10 +117,10 @@
     <section id="url">
         <h2>URI Schemes</h2>
         <p>
-            CoAP uses the following URI schemes for identifying CoAP resources
+            CoAP uses the following Uniform Resource Identifier (URI) schemes for identifying CoAP resources
             and providing a means of locating the resource:
         </p>
-        <table class="def">
+        <table class="data">
             <thead>
                 <tr>
                     <th>URI Scheme</th>
@@ -155,8 +155,8 @@
             </tbody>
         </table>
         <p>
-            See <a href="https://www.iana.org/assignments/uri-schemes/">Uniform Resource Identifier (URI) Schemes</a>
-            for the complete list of CoAP URI schemes.
+            See the <a href="https://www.iana.org/assignments/uri-schemes/">Uniform Resource Identifier (URI) Schemes</a>
+            registry for the complete list of CoAP URI schemes.
         </p>
     </section>
     <section id="vocabulary">
@@ -229,10 +229,10 @@
                         <td><a href="#BlockWiseTransferParameters">BlockWiseTransferParameters</a></td>
                     </tr>
                     <tr>
-                        <td><code>coap:quickblockwise</code></td>
+                        <td><code>coap:qblockwise</code></td>
                         <td>Indicates that a quick block-wise transfer [[RFC9177]] is supported</td>
                         <td>optional</td>
-                        <td><a href="#BlockWiseTransferParameters">QuickBlockWiseTransferParameters</a></td>
+                        <td><a href="#BlockWiseTransferParameters">QBlockWiseTransferParameters</a></td>
                     </tr>
                     <tr>
                         <td><code>coap:hopLimit</code></td>
@@ -278,8 +278,8 @@
                 </tbody>
             </table>
         </section>
-        <section id="QuickBlockWiseTransferParameters">
-            <h3>QuickBlockWiseTransferParameters</h3>
+        <section id="QBlockWiseTransferParameters">
+            <h3>QBlockWiseTransferParameters</h3>
             <table class="def">
                 <thead>
                     <tr>
--- a/bindings/protocols/coap/index.html
+++ b/bindings/protocols/coap/index.html
@@ -16,22 +16,9 @@
             edDraftURI: "https://w3c.github.io/wot-binding-templates/bindings/protocols/coap/",
             shortName: "wot-coap-binding",
             github: "https://github.com/w3c/wot-binding-templates/tree/main/bindings/protocols/coap",
-            localBiblio: {
-                "I-D.ietf-core-new-block": {
-                    title: "Constrained Application Protocol (CoAP) Block-Wise Transfer Options Supporting Robust Transmission",
-                    href: "https://datatracker.ietf.org/doc/draft-ietf-core-new-block/14/",
-                    "status": "Internet Draft (work in progress)",
-                    "publisher": "IETF",
-                    "date": "14 February 2022",
-                    "authors": [
-                        "M. Boucadair",
-                        "J. Shallow"
-                    ],
-                },
-            },
         };
     </script>
-    <title>CoAP binding template</title>
+    <title>CoAP Binding Template</title>
 </head>
 
 <body>
@@ -114,6 +101,7 @@
                     <td>Hop Limit</td>
                     <td><code>"coap:hoplimit": ...</code></td>
                 </tr>
+                <!--
                 <tr>
                     <td>OSCORE</td>
                     <td><code>SecurityScheme</code></td>
@@ -122,6 +110,7 @@
                     <td>EDHOC</td>
                     <td><code>SecurityScheme</code></td>
                 </tr>
+                -->
             </tbody>
         </table>
     </section>
@@ -167,7 +156,7 @@
         </table>
         <p>
             See <a href="https://www.iana.org/assignments/uri-schemes/">Uniform Resource Identifier (URI) Schemes</a>
-            for a complete list of CoAP URI schemes.
+            for the complete list of CoAP URI schemes.
         </p>
     </section>
     <section id="vocabulary">
@@ -194,29 +183,54 @@
                         <td><code>coap:method</code></td>
                         <td>Indicates the CoAP method to be used in the request</td>
                         <td>optional</td>
-                        <td><a href="#Method">Method</a></td>
+                        <td>
+                            <a href="https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#string"><code>string</code></a>
+                            <p>
+                                (see the
+                                <a href="https://www.iana.org/assignments/core-parameters/#method-codes">CoAP Method Codes</a>
+                                registry for the set of possible values)
+                            </p>
+                        </td>
                     </tr>
                     <tr>
                         <td><code>coap:format</code></td>
-                        <td>Indicates that a payload and a CoAP Content-Format option [[RFC7252]] be included in the request</td>
+                        <td>Indicates that a payload in one of the specified formats with a matching CoAP Content-Format option [[RFC7252]] be included in the request</td>
                         <td>optional</td>
-                        <td><a href="#ContentFormat">ContentFormat</a> or array of <a href="#ContentFormat">ContentFormat</a></td>
+                        <td>
+                            <a href="https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#unsignedShort"><code>unsignedShort</code></a>
+                            or array of
+                            <a href="https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#unsignedShort"><code>unsignedShort</code></a>
+                            <p>
+                                (see the
+                                <a href="https://www.iana.org/assignments/core-parameters/#content-formats">CoAP Content-Formats</a>
+                                registry for the set of possible values)
+                            </p>
+                        </td>
                     </tr>
                     <tr>
                         <td><code>coap:accept</code></td>
-                        <td>Indicates that a CoAP Accept option [[RFC7252]] be included in the request</td>
+                        <td>Indicates that a CoAP Accept option [[RFC7252]] with one of the specified formats be included in the request</td>
                         <td>optional</td>
-                        <td><a href="#ContentFormat">ContentFormat</a></td>
+                        <td>
+                            <a href="https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#unsignedShort"><code>unsignedShort</code></a>
+                            or array of
+                            <a href="https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#unsignedShort"><code>unsignedShort</code></a>
+                            <p>
+                                (see the
+                                <a href="https://www.iana.org/assignments/core-parameters/#content-formats">CoAP Content-Formats</a>
+                                registry for the set of possible values)
+                            </p>
+                        </td>
                     </tr>
                     <tr>
                         <td><code>coap:blockwise</code></td>
-                        <td>Indicates that a block-wise transfer is supported [[RFC7959]]</td>
+                        <td>Indicates that a block-wise transfer [[RFC7959]] is supported</td>
                         <td>optional</td>
                         <td><a href="#BlockWiseTransferParameters">BlockWiseTransferParameters</a></td>
                     </tr>
                     <tr>
                         <td><code>coap:quickblockwise</code></td>
-                        <td>Indicates that a quick block-wise transfer is supported [[I-D.ietf-core-new-block]]</td>
+                        <td>Indicates that a quick block-wise transfer [[RFC9177]] is supported</td>
                         <td>optional</td>
                         <td><a href="#BlockWiseTransferParameters">QuickBlockWiseTransferParameters</a></td>
                     </tr>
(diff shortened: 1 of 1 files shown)


bindings/protocols/coap/index.html
```diff
@@ -1,181 +1,464 @@
...
+                    "authors": [
+                        "M. Boucadair",
+                        "J. Shallow"
+                    ],
+                },
+            },
         };
     </script>
-    <title>CoAP Binding Template</title>
+    <title>CoAP binding template</title>
```
egekorkan, line 34: It would be better to keep capitilization, I have realized that it is wrong in Modbus now

egekorkan (changes requested):
Before you make further changes, I would also recommend to work with the toolchain approach. At least doing the edits in the template.html file would make it easier

bindings/protocols/coap/index.html
```diff
@@ -1,181 +1,464 @@
...
+                    <td><code>"coap:hoplimit": ...</code></td>
+                </tr>
+                <tr>
+                    <td>OSCORE</td>
+                    <td><code>SecurityScheme</code></td>
+                </tr>
+                <tr>
+                    <td>EDHOC</td>
+                    <td><code>SecurityScheme</code></td>
+                </tr>
```
JKRhb, line 124: Can we define these Schemes in this document? Or is a rework of the process for registering SecuritySchemes needed to be able to so? (This is probably also related to https://github.com/w3c/wot-thing-description/issues/1395.)

bindings/protocols/coap/index.html
```diff
@@ -1,181 +1,464 @@
...
+                        <td><code>coap:method</code></td>
+                        <td>Indicates the CoAP method to be used in the request</td>
+                        <td>optional</td>
+                        <td><a href="#Method">Method</a></td>
+                    </tr>
+                    <tr>
+                        <td><code>coap:format</code></td>
+                        <td>Indicates that a payload and a CoAP Content-Format option [[RFC7252]] be included in the request</td>
+                        <td>optional</td>
+                        <td><a href="#ContentFormat">ContentFormat</a> or array of <a href="#ContentFormat">ContentFormat</a></td>
```
JKRhb, line 203: According to [RFC 7252 section 5.10](https://datatracker.ietf.org/doc/html/rfc7252#section-5.10), the Content-Format option is not repeatable. Is it intentional to allow an array of Content-Formats here?

bindings/protocols/coap/index.html
```diff
@@ -1,181 +1,464 @@
...
+                    <td>Conditional Requests</td>
+                    <td>n/a</td>
+                </tr>
+                <tr>
+                    <td>Content</td>
+                    <td><code>"coap:format": ...</code></td>
+                </tr>
+                <tr>
+                    <td>Content Negotiation</td>
+                    <td><code>"coap:accept": ...</code></td>
```
JKRhb, line 99: As discussed in #49, `coap:format` and `coap:accept` could also be implied by the `contentType` and/or the [`response`](https://w3c.github.io/wot-thing-description/#td-vocab-response--Form) fields.

JKRhb (commented):
Thank you for starting this PR, @ektrah! See a couple of first comments below :)

bindings/protocols/coap/index.html
ektrah: I've started off with index.template.html and just replaced `BINDINGNAME` with `CoAP`:

https://github.com/w3c/wot-binding-templates/blob/09de2a43d80188171c9d265de4d17d6444382066/bindings/index.template.html#L20

Will fix.

bindings/protocols/coap/index.html
```diff
@@ -1,181 +1,464 @@
...
+                    <td>Caching: Validation</td>
+                    <td>n/a</td>
+                </tr>
+                <tr>
+                    <td>Conditional Requests</td>
+                    <td>n/a</td>
+                </tr>
+                <tr>
+                    <td>Content</td>
+                    <td><code>"coap:format": ...</code></td>
```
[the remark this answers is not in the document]
ektrah: Since it looks like we're going to change quite a bit, I'm wondering if we shouldn't rather intentionally change the prefix...

bindings/protocols/coap/index.html
ektrah: Good question.

bindings/protocols/coap/index.html
ektrah: The thought was that a server actually might accept multiple content-formats (e.g., both SenML/CBOR and SenML/JSON), which could be listed in the TD (probably in the order of the server's preference). A client would pick the most appropriate one and include a payload (and Content-Format option) accordingly.

bindings/protocols/coap/index.html
ektrah: True. I think we need to discuss our assumptions regarding a client's knowledge of content types and CoAP content formats: Do we assume that a client is capable of translating a string such as "text/plain; charset=utf-8" to the ID 0 (which would pretty much require that the client has a copy of the IANA registry) or would it be easier if the ID was included in the TD right away?

bindings/protocols/coap/index.html
ektrah: BTW, why are these documents called *templates*? Would perhaps "CoAP Binding for Web of Things" be a better title?

bindings/protocols/coap/index.html
```diff
@@ -1,181 +1,464 @@
...
+                    </tr>
+                </tbody>
+            </table>
+        </section>
+        <section id="Method">
+            <h3>Method</h3>
+            <p>
+                ... including (but not limited to) the following methods:
+            </p>
+            <table class="def" title="Method">
```
ektrah, line 237: Since this table is essentially a copy of the IANA registry, should we have it at all?

bindings/protocols/coap/index.html
```diff
@@ -1,181 +1,464 @@
...
+                <a href="https://www.iana.org/assignments/core-parameters/#method-codes">Constrained RESTful Environments (CoRE) Parameters</a>
+                for a complete list of methods.
+            </p>
+        </section>
+        <section id="ContentFormat">
+            <h3>ContentFormat</h3>
+            <p>
+                ... including (but not limited to) the following content formats:
+            </p>
+            <table class="def" title="ContentFormat">
```
ektrah, line 286: Since this table is essentially a copy of the IANA registry, should we have it at all?

bindings/protocols/coap/index.html
```diff
@@ -1,181 +1,464 @@
...
+            <table class="def">
+                <thead>
+                    <tr>
+                        <th>Operation</th>
+                        <th>Default Binding</th>
+                    </tr>
+                </thead>
+                <tbody>
+                    <tr>
+                        <td><code>td:readProperty</code></td>
```
ektrah, line 379: Should this be `<td><code>readproperty</code></td>`?

bindings/protocols/coap/index.html
```diff
@@ -1,181 +1,464 @@
...
+                        <td><code>"coap:method": "coap:get"</code></td>
```
ektrah, line 380: Should this be `<td><code>"coap:method": "get"</code></td>`?

bindings/protocols/coap/index.html
ektrah: Re-reading [the section on `Form`](https://w3c.github.io/wot-thing-description/#form), it's also a bit unclear to me what `contentType` and `contentCoding` actually indicate. Do these describe the format of the payload that the client is supposed to submit (&rarr; Content-Format option in CoAP request), the format of the response payload to expect (&rarr; Content-Format option in CoAP response), or the format of the response payload that the client is supposed to request (&rarr; Accept option in CoAP request)?

JKRhb: I found that a bit unclear as well. My latest understanding was that `contentType` describes the payload format both of requests and responses unless there is a response field set which only refers to the response. The spec also states:

> The optional response name-value pair can be used to provide metadata for the expected response message. [...] If no response name-value pair is provided, it MUST be assumed that the content type of the response is equal to the content type assigned to the Form instance.

So if both are set, I think you could argue that `contentType` should refer to the Content-Format option and `response.contentType` should refer to the `Accept` option. If only the Form `contentType` is given, then it would refer to both options. I am not sure if this is the correct interpretation. Maybe some clarification in the TD spec would be required. 

node-wot is currently using the contentType only for the acccept option by the way: https://github.com/eclipse/thingweb.node-wot/blob/master/packages/binding-coap/src/coap-client.ts#L256 The Content-Format needs to be explicitly set as an option in the Form.

> I think we need to discuss our assumptions regarding a client's knowledge of content types and CoAP content formats: Do we assume that a client is capable of translating a string such as "text/plain; charset=utf-8" to the numeric ID 0 (which would pretty much require that the client has a copy of the IANA registry) or would it be easier if the numeric ID was included in the TD right away?

That is a very good question! I think the mapping between string-based contentTypes and CoAP's Content-Format is also a bit difficult in general if you have (additional) parameters. Being able to use the value directly is certainly better, especially for consumers that are more constrained. Maybe for TD 2.0 it could also be allowed to use an integer for `contentType`? In this case the protocol specific meaning could be inferred from the `href`'s URI scheme. Or would it be better to keep it only for human-readable information?

In general, I think there should probably be some kind of recommendation/(pseudo-)assertion that the values have to match as given by the IANA registry if both `coap:format` and `contentType` are being used.

bindings/protocols/coap/index.html
JKRhb: I think a reference to IANA would probably be the most elegant solution. Maybe an example for the use of the option values could be given instead?

bindings/protocols/coap/index.html
JKRhb: Yes, the operation type values are all lowercase (see https://w3c.github.io/wot-thing-description/#table-operation-types).

Should the operation types be prefixed by the way? In the HTTP binding the `td:` prefix is missing, for example: https://w3c.github.io/wot-binding-templates/bindings/protocols/http/index.html#http-default-vocabulary-terms

bindings/protocols/coap/index.html
JKRhb: In the HTTP binding and the old CoAP binding, the method names are unprefixed and all upper case, i. e.:

```suggestion
                        <td><code>"coap:method": "GET"</code></td>
```

JKRhb: Or would it make sense to also prefix the method names from an RDF standpoint?

bindings/protocols/coap/index.html
JKRhb: Oh, I see. This would also be consistent with the `ct` link parameter from the CoRE RD draft, so I think that sounds good :)

bindings/protocols/coap/index.html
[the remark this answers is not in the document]
JKRhb: Good point ;) It seems as if there was the time when this was already the case, though: https://github.com/w3c/wot-binding-templates/blob/e4fd7a4f925b7da063c6f99f9a1debb721a22d65/index.html I couldn't find the reason why the prefix was changed from `coap:` to `cov:` (probably because `http:` was changed to `htv:` in order to prevent name clashes...?) but I would also be in favor of using `coap:` to make it more clear which protocol is being referred to.

egekorkan:
Call of 02.03:
- We will wait for @ektrah for the discussions
- Issues can be created after merge of the PR to keep on discussing but discussing here is fine for now

bindings/protocols/coap/index.html
egekorkan: That is a valid discussion, also see https://github.com/w3c/wot-binding-templates/issues/143

bindings/protocols/coap/index.html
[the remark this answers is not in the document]
egekorkan: this is a rather long discussion:
- https://github.com/w3c/wot-thing-description/issues/443
- In the modbus PR from @relu91 we had a discussion on this but I could not find it

bindings/protocols/coap/index.html
egekorkan: I would propose to not add these as new terms but a sort of hint on these available in TD 2.0. TD security mechanism needs rework to allow this in my opinion

bindings/protocols/coap/index.html
egekorkan: No need for the prefix, why should they be prefixed? If you mean the ontology, the TD ontology is used without a prefix so all the keywords can (I think MUST) be used without a prefix.

bindings/protocols/coap/index.html
egekorkan: Yes, the prefix can be omitted. The reason for using capital letters is that the HTTP in RDF used these: https://www.w3.org/TR/HTTP-in-RDF10/#example

bindings/protocols/coap/index.html
egekorkan: > I think we need to discuss our assumptions regarding a client's knowledge of content types and CoAP content formats: Do we assume that a client is capable of translating a string such as "text/plain; charset=utf-8" to the numeric ID 0 (which would pretty much require that the client has a copy of the IANA registry) or would it be easier if the numeric ID was included in the TD right away?

This was also a discussion for MQTT where we could have used the integer values that are actually put into the header instead of `"SUBSCRIBE"` for example. I have then looked at some libraries and none of them actually expose these integer values to the programming API. Thus, we have chosen the words instead of the numbers. Here I would recommend the same thing: Go the way most libraries' APIs have gone

bindings/protocols/coap/index.html
```diff
@@ -1,181 +1,464 @@
...
+                <tr>
+                    <td>coaps</td>
+                    <td>[[RFC7252]]</td>
+                </tr>
+                <tr>
+                    <td>coaps+tcp</td>
+                    <td>[[RFC8323]]</td>
+                </tr>
+                <tr>
+                    <td>coaps+ws</td>
```
egekorkan, line 163: for coap+ws as well

Should this instead be a subprotocol and the href should be ws ? This is a real question btw :)

JKRhb: `coap+ws` and `coaps+ws` are actually official URI schemes, [specified in RFC 8323](https://datatracker.ietf.org/doc/html/rfc8323#section-8.3) :) So I think it is probably best to keep it in the href.

JKRhb:
How do you think we should proceed with this PR? As I see it there are a couple of comments which could be resolved here while a few others might be require some discussion in the Binding Templates call. Can we maybe get this PR to a mergeable state in the next couple of weeks?

bindings/protocols/coap/index.html
ektrah: https://www.w3.org/2022/wot/td/v1.1 says:

```
        "readproperty": "td:readProperty",
```

So it seems that, in the ontology, it's *not* all lowercase, and you need to use the prefix if you want that spelling...

I'll change it to `readproperty`.

bindings/protocols/coap/index.html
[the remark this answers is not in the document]
ektrah: > What is format by the way? I find different meanings when looking at Wikipedia and rfc

A CoAP content format is the combination of an HTTP content type (such as `text/plain;charset=utf-8`) with an HTTP content coding (such as `deflate`). The CoAP Content-Format Option describes the content format of the payload in a request or response. The CoAP Accept Option in a request specifies the content format of the payload in a successful response.

ektrah:
I've just pushed a commit that (hopefully adequately) addresses the low-hanging fruits. There are some open issues that would benefit from some discussion in the Binding Templates call:

* Document title (#143)
* Choice of prefix (`coap:`, `cov:`, or something else)
* Unclear semantics of `contentType` and `contentCoding`
* CoAP content formats expressed by strings (such as `"text/plain;charset=utf-8"` and `"deflate"`) or numeric identifiers

However, if there is general agreement that this is overall the right approach, we could merge now/after another quick round of reviews and address these open issues afterwards.

egekorkan:
Call of 06.04:
- The group agrees on the new feature based approach, thus this PR can be merged
- The four points mentioned above will be addressed in separate (sometimes new) issues

PR #150, opened 2022-03-01
Add pre-commit hook for rendering documentation

relu91:
I've set up a `husky` pre-commit hook to force the `node render.js` before sending changes to the main repository. I thought it might help prevent the pains that we have every time we review PRs that forget to run the `renderer`. As a side effect, the installation of husky updated the package-lock to the latest version 😢 .


egekorkan:
This would be also valid for the TD repo right? :eyes:

egekorkan (approved):

PR #156, opened 2022-04-08
Improve the `mqtt` protocol binding template

relu91:
This PR improves the organization of the `mqtt` protocol binding template and its ontology. The `.ttl` file has been cleaned up and adapted to the Protégé file format. The ontology classes now follow better the [MQTT v3.1.1](https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/mqtt-v3.1.1.html) specification with dedicated entities for `filter`s and `topic`s.


egekorkan:
@relu91 is this ready for review?

relu91:
> @relu91 is this ready for review?

Yes!

egekorkan:
Postponed for next week, a barebones preview is available at https://htmlpreview.github.io/?https://raw.githubusercontent.com/relu91/wot-binding-templates/mqtt-cont/bindings/protocols/mqtt/index.html

egekorkan (changes requested):
Note: More to come

Additionally:
- I really like that there are granular examples for each vocabulary term. It would be even nicer to link to examples from the sections :)

egekorkan:
One important thing that is bothering me is the use of `href` in combination with `mqv:topic` and `mqv:filter`. The `href` is clearly used as the broker URL in the examples and does not contain the topic names to publish or to subscribe to. This is however not mentioned clearly. Additionally, the `mqv:topic` and `mqv:filter` are both optional but one of them has to be present to identify the topic(s). Also, the `mqv:topic` and `mqv:filter` are not explained thorough enough to understand the difference. In the MQTT world, topic is always a singular thing that message always contains. The filter is sent in subscription requests. Here we have to say what means to have filter or topic in a subscription or publish message. Ideally, subscription should be filter and publish should be topic but publish can also have a filter to say that the Consumer can publish to any of the matching topics. 

I propose to have a small discussion on this synchronously.

bindings/protocols/mqtt/index.template.html
```diff
@@ -173,9 +188,215 @@ <h3>Possible mappings</h3>
   </section>
     <section>
       <h2>Examples</h2>
-      <p>The following examples show how to use the MQTT protocol in a Thing Description.</p>
-      <p class="ednote">TODO:</p>
+      <p>
+        The following examples show how to use the MQTT protocol in a Thing Description. The [[[#example-simple-event]]] 
+        shows a simple form that indicates to compliant clients to subscribe or unsubscribe to the topic <code>thing1/events/overheating</code>.
```
egekorkan, line 193:
```suggestion
        shows a simple form that indicates to compliant Consumers to subscribe or unsubscribe to the topic <code>thing1/events/overheating</code> in order to execute the <code>subscribevent</code> or <code>unsubscribevent</code> operation, respectively.
```

bindings/protocols/mqtt/index.template.html
```diff
@@ -173,9 +188,215 @@ <h3>Possible mappings</h3>
...
+            "href": "mqtt://broker.com:1883",
+            "op": [
+                "invokeaction",
+            ], 
+            "mqv:topic": "application/devices/thing1/program/commands/reset",
+        }
+      </pre>
+      <p>
+        Continuing with increasingly more complex examples, [[[#example-property]]] shows how to use the MQTT protocol when handling properties.
+        <code>mqv:retain</code> is used to indicate the ability to read the property rather than just observe it.
```
egekorkan, line 221: I think here it is needed to mention that retain only applies to reading a property. I think this example can be very confusing since it actually means that the Consumer needs to set retain to true for writing. Once I think about it, we do not need to put retain information in TDs since having readproperty implies that. It is not something that the Consumer can really use by reading the TD. It is good for giving a hint that this topic will contain a retained message that is the property value but not more.

bindings/protocols/mqtt/index.template.html
```diff
@@ -173,9 +188,215 @@ <h3>Possible mappings</h3>
...
+                "writeproperty",
+                "observeproperty"
+            ], 
+            "mqv:retain" : true,
+            "mqv:topic": "application/devices/thing1/properties/test",
+            "mqv:filter": "application/devices/thing1/properties/test",
+        }
+      </pre>
+      <p>
+        If the implementer want to give the ability to WoT clients to conveniently observe all properties in a single call, it is possible
```
egekorkan, line 237:
```suggestion
        If the implementer wants to give the ability to WoT clients to conveniently observe all properties in a single call, it is possible
```

bindings/protocols/mqtt/index.template.html
```diff
@@ -173,9 +188,215 @@ <h3>Possible mappings</h3>
...
+        to use observerallproperties form in the root of a Thing Description. 
+      </p>
+      <pre id="example-property" class="example" title="Read a single coil">
```
egekorkan, line 240: the id is duplicate

bindings/protocols/mqtt/index.template.html
```diff
@@ -173,9 +188,215 @@ <h3>Possible mappings</h3>
...
+                "observeproperty"
+            ], 
+            "mqv:retain" : true,
+            "mqv:topic": "application/devices/thing1/properties/test",
+            "mqv:filter": "application/devices/thing1/properties/test",
+        }
+      </pre>
+      <p>
+        If the implementer want to give the ability to WoT clients to conveniently observe all properties in a single call, it is possible
+        to use observerallproperties form in the root of a Thing Description.
```
egekorkan, line 238: I think it would be nice to mention that this means all messages published to the topics under `application/devices/thing1/properties/` will be delivered.

bindings/protocols/mqtt/index.template.html
```diff
@@ -173,9 +188,215 @@ <h3>Possible mappings</h3>
...
+                      "observeproperty"
+                  ], 
+                  "mqv:qos": "qos:1", 
+                  "mqv:retain" : true,
+                  "mqv:topic": "application/devices/thing1/properties/test",
+                  "mqv:filter": "application/devices/thing1/properties/test",
+              }
+     </pre>
+      <p>
+        Finally, [[[#example-full]]] reports a full Thing Description for a commercial device. The broker address is only for
```
egekorkan, line 269: commercial device is a bit weird to use here. It makes me think of an already existing product

bindings/protocols/mqtt/index.template.html
```diff
@@ -173,9 +188,215 @@ <h3>Possible mappings</h3>
...
+                "unknown",
+                "warmup",
+                "normal",
+                "fault"
+            ],
+            "type": "string",
+            "forms": [
+                {
+                    "href": "mqtt://broker.com/",
+                    "mqtt:filter": "application/{{deviceid}}/sensor/operation",
```
egekorkan, line 299: Is this supposed to be a TM?

egekorkan (changes requested):
Also, there is no bibliography

bindings/protocols/mqtt/index.template.html
```diff
@@ -173,9 +188,215 @@ <h3>Possible mappings</h3>
...
+                  "mqv:retain" : true,
+                  "mqv:topic": "application/devices/thing1/properties/test",
+                  "mqv:filter": "application/devices/thing1/properties/test",
+              }
+     </pre>
+      <p>
+        Finally, [[[#example-full]]] reports a full Thing Description for a commercial device. The broker address is only for
+        demonstration purposes. 
+      </p>
+      <pre id="example-full" class="example" title="Read a single coil">
```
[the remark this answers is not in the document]
relu91: Not yet resolved, let's discuss today

relu91:
Notice that I found more terms from  [MQTT v3.1.1](https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/mqtt-v3.1.1.html) spec that can be described inside our ontology, like Application Messages and Subscriptions. I would defer these changes later since they only effect the correctness of the protocol ontology but not the protocol binding template.

PR #160, opened 2022-05-25
Add license and coc

egekorkan:
I am adding these since I have added this to Netlify as well. As noted at https://github.com/w3c/wot-marketing/issues/304

The website is at https://wot-binding-templates.netlify.app/

**Note:** We are using netlify only for PR reviews

### DIFF
--- a/LICENSE copy.md
+++ b/LICENSE copy.md
@@ -1,0 +1,9 @@
+All Reports in this Repository are licensed by Contributors
+under the
+[W3C Software and Document License](http://www.w3.org/Consortium/Legal/2015/copyright-software-and-document).
+
+Contributions to Specifications are made under the
+[W3C CLA](https://www.w3.org/community/about/agreements/cla/).
+
+Contributions to Test Suites are made under the
+[W3C 3-clause BSD License](https://www.w3.org/Consortium/Legal/2008/03-bsd-license.html)
--- a/LICENSE copy.md
+++ b/LICENSE copy.md
@@ -1,9 +1,0 @@
-All Reports in this Repository are licensed by Contributors
-under the
-[W3C Software and Document License](http://www.w3.org/Consortium/Legal/2015/copyright-software-and-document).
-
-Contributions to Specifications are made under the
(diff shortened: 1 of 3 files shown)


LICENSE copy.md
```diff
@@ -0,0 +1,9 @@
+All Reports in this Repository are licensed by Contributors
```
danielpeintner, line 1: what is the right term?
* reports
* documents
* ?

LICENSE copy.md
```diff
@@ -0,0 +1,9 @@
...
+under the
+[W3C Software and Document License](http://www.w3.org/Consortium/Legal/2015/copyright-software-and-document).
+
+Contributions to Specifications are made under the
+[W3C CLA](https://www.w3.org/community/about/agreements/cla/).
+
+Contributions to Test Suites are made under the
+[W3C 3-clause BSD License](https://www.w3.org/Consortium/Legal/2008/03-bsd-license.html)
```
danielpeintner, line 9: I am not sure whether those to items apply, do they?
I TD [LICENSE.md](https://github.com/w3c/wot-thing-description/blob/main/LICENSE.md) it is not there.

danielpeintner (changes requested):
I think you used the wrong file name `LICENSE copy.md`

Some more comments below.

LICENSE copy.md
egekorkan: Reports includes notes and recommendations. This is something I have copied from W3C and would not want to change anything

LICENSE copy.md
egekorkan: Not sure why TD does not have this information but my guess would that its license is simply older than what W3C asks for now

egekorkan:
@danielpeintner I have changed the filename, it was (obviously ^^) a copy paste mistake. The actual content of the license and CoC is something I do not want to change since they come from W3C

relu91 (approved):

danielpeintner (approved):

PR #177, opened 2022-09-07
fix link errors to ontology

### LINKED ISSUE #166: [MQTT] fixing link errors
The MQTT Vocabulary section (https://w3c.github.io/wot-binding-templates/bindings/protocols/mqtt/index.html#mqtt-vocabulary) has broken links since they need to link to the ontology document

egekorkan:
fixes #166 

However, the ontology document does not use relative links anymore. If someone reads the ontology document, the links like `TopicFilter` will load a new page.

I do not know how to fix it 100% without changing the rendering script but I think this is a definitive improvement


relu91 (commented):
Yes, I remember this issue. It is a discrepancy in how the spaces are treated in the two renders (ontology and binding). I would prefer to harmonize the logic rather than hard code the link, but yes this for sure is an improvement. Let you decide whether to fix this later or in this PR

egekorkan:
I will create another issue to fix that separately in another PR

PR #181, opened 2022-09-07
Fix examples, adding payload-driven usage of mqtt

### LINKED ISSUE #167: MQTT binding examples
small changes
- example 7 has some small json errors (comma before `}` ) 
-  also it would be good to use different topics for read/observe and write in the example, since this is very often done in mqtt applications / topics layouts

the text and examples do not touch on the content problem of mqtt .. that everybody is crafting their own message payload.
the examples seem to assume a "every property their own subtopic, content as json", which is only one of several ones I've seen.  E.g. also the ""every property their own subtopic, content as string (i.e. no quotes and json escapes)" and "every thing their own subtopic, content as json".

I understand, if this would at current time overload the mqtt binding (I'd rather have a version released now and extend later), but it would help if this is made explicit as "out-of-scope" or "out-of-scope for now".

egekorkan:
fixes #167 

I have 
- corrected some capitalization issues
- made sure that all examples are valid JSON. There were extra commas in many lines
- added explanation that a payload over MQTT is possible where affordances would have same topics

### DIFF
--- a/bindings/protocols/mqtt/index.template.html
+++ b/bindings/protocols/mqtt/index.template.html
@@ -61,10 +61,10 @@
 <body>
     <section id="abstract">
       In the context of the Web of Things, a Binding Template is 
-      a blueprint that gives guidance on how to implement a specific IoT protocol, data format or IoT platform. This document 
-      gives implementation guidelines regarding the MQTT protocol. The MQTT protocol is
-      a lightweight, asynchronous, transport-independent, open protocol for publishing and subscribing to messages. It has 
-      been employed in many IoT platforms and IoT applications for its simplicity and scalability.
+      a blueprint that gives guidance on how to implement a specific IoT protocol, data format or IoT platform. 
+      This document gives implementation guidelines regarding the MQTT protocol. 
+      The MQTT protocol is a lightweight, asynchronous, transport-independent, open protocol for publishing and subscribing to messages. 
+      It has been employed in many IoT platforms and IoT applications for its simplicity and scalability.
     </section>
     <section id='introduction'>
         <h2>Introduction</h2>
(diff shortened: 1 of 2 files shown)


egekorkan:
@a-hennig please have a look :)

bindings/protocols/mqtt/index.template.html
```diff
@@ -75,20 +75,27 @@ <h2>Introduction</h2>
...
-            
+        </p>
+        <p>
             In the context of the Internet of Things, many vendors expose MQTT publishers mixed with 
             Web protocols like HTTP. This document explain how to describe MQTT endpoints and payloads with 
             Thing Descriptions (TDs) the core building block of the Web of Things (WoT) [[WOT-ARCHITECTURE]].
         </p>
+        <p>
+            This version of the document is designed around each affordance having its own MQTT topic.
+            Usage of MQTT with an additional protocol on top that uses topics with specific payloads is possible but it
```
JKRhb, line 86:
```suggestion
            Usage of MQTT with an additional protocol on top that uses topics with specific payloads is possible, but it
```

bindings/protocols/mqtt/index.template.html
```diff
@@ -111,10 +118,9 @@ <h2>URL format</h2>
                 <li><code>{address}</code> is the IP address of the MQTT broker</li>
                 <li><code>{port}</code> is the port of the MQTT broker</li>
             </ul>
-            Notice that the URL does not contain any information regarding [[[#topicname]]] or [[[#topicfilter]]]. These two
-            parameters
-            must to be specified using their own parameter as described in [[[#form-terms]]]. This to avoid ambiguities
-            regarding the starting <code>/</code> character and the usage of <code>#</code> that can be confused with URL
+            Notice that the URL does not contain any information regarding [[[#topicname]]] or [[[#topicfilter]]]. 
+            These two parameters must to be specified using their own parameter as described in [[[#form-terms]]].
```
JKRhb, line 122:
```suggestion
            These two parameters must be specified using their own parameter as described in [[[#form-terms]]]. 
```

bindings/protocols/mqtt/index.template.html
```diff
@@ -111,10 +118,9 @@ <h2>URL format</h2>
                 <li><code>{address}</code> is the IP address of the MQTT broker</li>
                 <li><code>{port}</code> is the port of the MQTT broker</li>
             </ul>
-            Notice that the URL does not contain any information regarding [[[#topicname]]] or [[[#topicfilter]]]. These two
-            parameters
-            must to be specified using their own parameter as described in [[[#form-terms]]]. This to avoid ambiguities
-            regarding the starting <code>/</code> character and the usage of <code>#</code> that can be confused with URL
+            Notice that the URL does not contain any information regarding [[[#topicname]]] or [[[#topicfilter]]]. 
+            These two parameters must to be specified using their own parameter as described in [[[#form-terms]]]. 
+            This to avoid ambiguities regarding the starting <code>/</code> character and the usage of <code>#</code> that can be confused with URL
```
JKRhb, line 123: Would it be more precise to say "URL fragment *separator*" here?

```suggestion
            This is necessary to avoid ambiguities regarding the starting <code>/</code> character and the usage of <code>#</code> which can be confused with the URL
```

bindings/protocols/mqtt/index.template.html
```diff
@@ -145,8 +151,8 @@ <h4>Form terms</h4>
       <h4>Control Packet</h4>
       <p>
         In MQTT protocol messages are called Control Packets. An MQTT Control Packet consists of up to three parts: fixed header (present in all Control Packets), variable header, payload. 
-          The fixed header consists of a packet type and flags. The main packet types that can be uses in a Form are listed below. Although, it is possible to use different
-        values of [[[#controlpacket]]] in forms, it is recommended to use the mapping defined in [[[#default-mappings]]].
+          The fixed header consists of a packet type and flags. The main packet types that can be uses in a Form are listed below. 
+          Although, it is possible to use different values of [[[#controlpacket]]] in forms, it is recommended to use the mapping defined in [[[#default-mappings]]].
```
JKRhb, line 155:
```suggestion
          Although it is possible to use different values of [[[#controlpacket]]] in forms, it is recommended to use the mapping defined in [[[#default-mappings]]].
```

JKRhb (commented):
See a couple of editorial suggestions below :)

egekorkan:
Thanks @JKRhb , I have added your suggestions. I just have one question that is linked at https://github.com/w3c/wot-binding-templates/pull/181#discussion_r968396579 but got lost when I committed your suggestion

bindings/protocols/mqtt/index.template.html
JKRhb: Oh, I actually meant the "fragment character" in the following line.

relu91 (approved):

PR #183, opened 2022-09-09
feat(modbus): move addres and quantity to URL components

### LINKED ISSUE #176: [Modbus] URI design for modbus+tcp URI schemes
I had a look at the Modbus binding and noticed that addressing information is split among URI and additional TD terms (e.g., `modbus:address`, `modbus:quantity`).

The URI should also for Modbus hold all the addressing information necessary to identify and locate the "Modbus resource" (i.e., registers, coils).

To illustrate: currently it is as if you would only put the "http://device.example.com:8080" part into the URI/href field, and define custom terms to hold "/api/properties/prop1"...

relu91:
This PR addresses the concerns raised in #176. Some possible discussion points:
- I stick with the familiar key=value syntax for the `quantity` parameter but in theory, we can shrink to just the number as we are defining the URL syntax ourselves
- I tried to keep the current document structure intact and just moved `unitID`, `address`, `quantity` to a dedicated section called URL terms. 
- I also added a very simple ABNF syntax for formally describing the components of the URL. Perhaps it is too much. 

Fixes #176

### DIFF
--- a/bindings/protocols/modbus/index.html
+++ b/bindings/protocols/modbus/index.html
@@ -70,7 +70,7 @@
         Considering that in Web of Things context all protocol bindings are required to at least support the Internet Protocol, for the Modbus protocol <code>modbus+tcp://</code> 
         was selected as the only valid URL scheme. The following shows the typical structure of an URL of the Modbus protocol:
         <pre>
-            modbus+tcp://{address}:{port}/{?unitID}
+            modbus+tcp://{address}:{port}/{unitID}/{address}?quantity={?quantity}
         </pre>
 
         <p>
@@ -79,13 +79,62 @@
                 <li><code>{address}</code> is the IP address of the Modbus device</li>
                 <li><code>{port}</code> is the port of the Modbus device</li>
                 <li><code>{unitID}</code> is the unit ID of the Modbus device. See <a href="#vocabulary">vocabulary</a></li>
+                <li><code>{address}</code> is the address of the register/coil referenced in this URL. See <a href="#vocabulary">vocabulary</a></li>
+                <li><code>{quantity}</code> the amount of registers/coils referenced in this URL. See <a href="#vocabulary">vocabulary</a></li>
             </ul>
+            For the full syntax see [[[#abnf]]]
         </p>
     </section>
     <section id="vocabulary">
         <h2>Modbus Vocabulary</h2>
         This section describes the vocabulary used in the Modbus protocol. A protocol binding implementation should use the vocabulary defined in this section to
         describe the different configuration that can be used to exchanged data between Web of Things.
+        <section>
+            <h3>URL terms</h3>
+            <table class="def">
+                <thead>
+                    <tr>
+                        <th>Vocabulary term</th>
+                        <th>Description</th>
+                        <th>Assignment</th>
+                        <th>Type</th>
+                    </tr>
+                </thead>
+                <tbody>
+                    <!-- Generate forms terms from the ontology-->
+                    
+        
+        <tr>
+            <td><code>modbus:address</code></td>
+            <td>Specifies the starting address of the Modbus operations</td>
+            <td>required</td>
+            <td><a href="http://www.w3.org/2001/XMLSchema#integer">integer</a></td>
+
+        <tr>
+        
+
+        
+        <tr>
+            <td><code>modbus:unitID</code></td>
+            <td>The Unit ID is usually not needed for ModbusTCP, since the IP-address works as unique identifier, but for compability reasons still often included</td>
+            <td>required</td>
+            <td><a href="http://www.w3.org/2001/XMLSchema#integer">integer</a></td>
+
+        <tr>
+        
+
+        
+        <tr>
+            <td><code>modbus:quantity</code></td>
+            <td>Specifies the amount of either registers or coils to be read or written to</td>
+            <td>optional</td>
+            <td><a href="http://www.w3.org/2001/XMLSchema#integer">integer</a></td>
+
+        <tr>
+        
+                </tbody>
+            </table>
+        </section>
         <section>
             <h3>Form terms</h3>
             <table class="def">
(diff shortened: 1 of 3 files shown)


egekorkan:
In the meeting of 12.10, there were multiple opinions:
- @sebastiankb prefers the more human readable option, which also would make sense in websocket subprotocols. Having two ways would be ok
- @danielpeintner says that we should decide on one method
- @relu91 the URL approach is better since it also allows usage of base to shorten the href
- @egekorkan talked with some colleagues. They agree with the philosophy behind the URL approach however the human readability is a big win for human readers.

@mkovatsc opinion is clear on the linked issue. 

We seem to be more in favor of the URL-based approach but not a full consensus. @mkovatsc can you join the TD+Binding call next week (4pm CEST)?

Also, if we change to the URL approach, we should reflect this in MQTT Binding as well.

egekorkan:
Some more relevant points from discussion with some colleagues:
- Validation and code completion are easier with vocabulary terms since JSON Schema can be used. Otherwise, we need to use regex or ABNF grammar.
- Implementation for consumer code seem to be not so important. Both methods are trivial to implement. Even implementing both is not a problem but it would be tricky to make sure that the TD does not contain both methods in a mixed way.
- (Also mentioned in the call of yesterday) With the URL approach a human TD reader or writer needs to open and read the binding spec each time. A compromise would be to put the vocabulary terms in the href like query parameters but this would result in a long uri.

bindings/protocols/modbus/index.template.html
```diff
@@ -337,13 +350,12 @@ <h2>Examples</h2>
         default for observing a property. 
         <pre  id="example-polling" class="example" title="Polling">
             {
-                "href": "modbus+tcp://127.0.0.1:60000/1",
+                "href": "modbus+tcp://127.0.0.1:60000/1/1",
                 "op": [
                     "observeproperty"
                 ],
                 "modbus:entity": "Coil",
                 "modbus:pollingTime": 1000,
```
mkovatsc, line 358: I am not sure a single device can define the polling time. For RTU it depends on on the number of devices on the cable and the baud rate. Hence, it should not be mobus vocabulary.

The underlying problem is independent from the transport. Didn't we have "stability" or "frequency" to indicate how often a value is expected to change?

bindings/protocols/modbus/index.template.html
```diff
@@ -390,11 +399,8 @@ <h2>Examples</h2>
                           "forms": [
                               {
                                   "op": "readproperty",
-                                  "href": "modbus+tcp://192.168.178.32:502",
+                                  "href": "./10002",
```
mkovatsc, line 402: see above

bindings/protocols/modbus/index.template.html
```diff
@@ -410,11 +416,8 @@ <h2>Examples</h2>
                                       "observeproperty",
                                       "readproperty"
                                   ],
-                                  "href": "modbus+tcp://192.168.178.32:502",
+                                  "href": "./6",
```
mkovatsc, line 419: see above

bindings/protocols/modbus/index.template.html
```diff
@@ -431,11 +434,8 @@ <h2>Examples</h2>
                                       "observeproperty",
                                       "readproperty"
                                   ],
-                                  "href": "modbus+tcp://192.168.178.32:502",
+                                  "href": "./3",
```
mkovatsc, line 437: see above

mkovatsc (changes requested):
Overall it address my original comment #176.

There are some things that need to be fixed (URI arithmetic).

I see the polling term in the modbus vocabulary as questionable; same problem applies for other transports without push mechanism.

We should also have support for Modbus RTU and possibly ASCII. However, it must be possible to easily see what kind of communication hardware would be required -- Modbus/TCP is just software, but the variants need special hardware. We will also need some global parameters, which is most likely the case for other bindings as well. A `bindingSchemes` structure analog to the `securitySchemes` structure could help for the global parameters as well as an explicit means to see what bindings a certain Thing would require.

egekorkan:
Opinions from the call of 19.10:
- We should collect the best practices to argue why we are doing it like this. There are already multiple implementations that use Modbus. node-wot has it and also @mkovatsc has implementations
- The URI scheme that we propose can contain the words to be more verbose.
- We should sync to all specs and use the same mechanism
- Tools: 
  - URLs are easy to copy compared to multiple terms since they are just strings
  - It is easier to treat resources (location, endpoints) as strings in an application. That way it is clear where the data came from or should go to
- We should look at the review by Matthias first. We will revisit it later but a decision is needed asap since it influences all subspecifications

bindings/protocols/modbus/index.html
```diff
@@ -70,7 +70,7 @@ <h2>URL format</h2>
         Considering that in Web of Things context all protocol bindings are required to at least support the Internet Protocol, for the Modbus protocol <code>modbus+tcp://</code> 
         was selected as the only valid URL scheme. The following shows the typical structure of an URL of the Modbus protocol:
         <pre>
-            modbus+tcp://{address}:{port}/{?unitID}
+            modbus+tcp://{address}:{port}/{unitID}/{address}?quantity={?quantity}
```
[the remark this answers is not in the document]
sebastiankb: I would prefer this version since it would meet

-  the human-readable aspect
-  meets the single address approach with href
-  would be clearer to define REG expressions for validation
-  would be clearer to define search queries (e.g., I like to find all properties that have quantity==4)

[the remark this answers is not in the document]
relu91: Although I agree that this version improves human-readability, I think it does not strictly comply with the semantics of URI components. In particular, I cite the definitions of [Path](https://www.rfc-editor.org/rfc/rfc3986#section-3.3) and [Query](https://www.rfc-editor.org/rfc/rfc3986#section-3.4) components from the RFC:

**Path**
> The path component contains data, usually organized in hierarchical
   form, that, along with data in the non-hierarchical query component
   ([Section 3.4](https://www.rfc-editor.org/rfc/rfc3986#section-3.4)), serves to identify a resource within the scope of the
   URI's scheme and naming authority (if any)
   
**Query**
> The query component contains non-hierarchical data that, along with
   data in the path component ([Section 3.3](https://www.rfc-editor.org/rfc/rfc3986#section-3.3)), serves to identify a
   resource within the scope of the URI's scheme and naming authority
   (if any).
   
In my understanding, the unitID and address properties are clearly hierarchical in nature as they define resources inside the space controlled by the base address. Moving this information in the query component will cause a loss of the real structure and it will mess with the URI arithmetic. In practise, I don't think we can define a common base address with the unitID anymore.

relu91:
I addressed the comments from @mkovatsc . Some other points:
> I see the polling term in the modbus vocabulary as questionable; same problem applies for other transports without push mechanism.

I agree, it came from node-wot implementation to give the ability to applications to use modbus also for observing properties. It is definitely a stretch and it is also hard to define the polling time in some use cases. However, I would discuss it in another issue and move on with this PR. 

> We should also have support for Modbus RTU and possibly ASCII. However, it must be possible to easily see what kind of communication hardware would be required -- Modbus/TCP is just software, but the variants need special hardware. We will also need some global parameters, which is most likely the case for other bindings as well

In my understanding, if we support Modbus RTU, it would be as if we support Bluetooth or I2C as it is a non-IP base protocol. In the use cases that I saw people use Modbus/TCP to use Modbus RTU or ASCII over the internet. If you feel like there something missing, I would like to discuss it in another issue :)

relu91:
> @relu91 I think we have agreed to merge this PR in the last week's meeting even though I do not see it at https://www.w3.org/2022/11/30-wot-td-minutes.html . We have created the issues https://github.com/w3c/wot-binding-templates/issues/214 and https://github.com/w3c/wot-binding-templates/issues/215 but they are not blockers for this. Any idea why we did not merge this one?

Not sure... I don't see any blockers right now. Maybe we opened the issues and forgot to merge this one?

egekorkan:
Call of 07.12: It seems we forgot to merge it last week. Merging

PR #188, opened 2022-09-29
Define CoAP Content Negotiation

### LINKED ISSUE #159: CoAP `Content-Format` vs. `contentType`/`contentCoding`
In the context of the new CoAP Binding Template, there were a couple of discussion (see e.g. https://github.com/w3c/wot-binding-templates/pull/149#discussion_r813994731) regarding the relationship between CoAP's `Content-Format`s which are numeric codes representing a combination of the `contentType` and `contentCoding` in TD forms.

While it is no problem mapping a `Content-Format` to a `contentType` and `contentCoding` (see the [corresponding table](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#content-formats) in the IANA registry), the other direction is not as straightforward as there are gaps in the set of Content-Types and encodings that are supported by CoAP (each new Content-Format has to be explicitly registered with IANA).

@ektrah's proposed solution in the current version of the CoAP Binding Template is to add a new vocabulary term called `coap:format`, where you can include numeric values for Content-Formats like so (also see #149 and #158):

```json
{
    "op": [ "writeproperty" ],
    "href": "coap://[2001:DB8::1]/.../...",
    "coap:method": "PUT",
    "coap:format": 112  // application/senml+cbor
}
```

This solves the problem that Consumers would need to always have a copy of the IANA registry otherwise.

However, there is the problem that forms have a default `contentType` of `application/json`, so in order to indicate a different Content-Format you would need to also override that field (c. f. https://github.com/w3c/wot-binding-templates/pull/158#discussion_r858922459):

```json
{
    "op": [ "writeproperty" ],
    "href": "coap://[2001:DB8::1]/.../...",
    "coap:method": "PUT",
    "coap:format": 112,
    "contentType": "application/senml+cbor"
}
```

For me, this raises a couple of questions/possible solutions for this problem in CoAP forms:

1. Should TDs be required to always indicate the `contentType` next to the `Content-Format` like in the second example? This has the disadvantage of creating redundancy, but would ensure maximum compatibility.
2. Show only "allow" `contentType` and `contentCoding` combinations be allowed that are mappable to CoAP Content-Formats? This would allow omitting/dropping the extra `coap:format` field, but Consumers would still need to have the IANA registry at hand.
3. (Should `contentType` and `contentCoding` be declared as HTTP-specific in TD 2.0? This could also address problems with other protocol binding templates.)

The last point is not suited for the current TD version, but should probably be kept in mind for the next one.

Looking forward to your thoughts and ideas on this issue!

ektrah:
Since the usage of `contentType` and `contentCoding` is really unclear and we don't seem to be making progress on this topic, I propose we just ditch that and define CoAP-specific vocabulary for content negotiation.

Closes #159

### DIFF
--- a/bindings/protocols/coap/index.html
+++ b/bindings/protocols/coap/index.html
@@ -100,7 +100,7 @@
             "forms": [{
                 "cov:method": "GET",
                 "href": "coap://[2001:DB8::1]/status",
-                "contentType": "text/plain;charset=utf-8",
+                "contentType": "",
                 "op": ["readproperty"]
             }]
         }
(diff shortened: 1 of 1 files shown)


JKRhb:
You are right that having specific vocabulary for the content format and accept options probably makes more sense than deriving them from the given combination of `contentType` and `contentCoding`. Setting `contentType` to a zero length string does not seem that elegant to me, though.

Since `contentType` does not need to be present in a form but has a default value of `application/json`, could we rather say that CoAP forms should either not assume the default at all or a Content-Format of `50` instead which could be overridden by `cov:available` and `cov:acceptable`?

ektrah:
(This PR relates to the discussion in https://github.com/w3c/wot-thing-description/pull/1564.)

ektrah:
Superseded by #193

PR #189, opened 2022-09-29
Delete "CoAP in RDF"

### LINKED ISSUE #97: CoAP RDF
From #93 

For today's kick-off meeting we should clarify following points:

1. Current status of the CoAP - RDF work (also see https://github.com/w3c/wot-binding-templates/tree/master/ontology)
2. Where will the work be managed (binding repo vs. new repo?)
3. Who coordinates the work?

ektrah:
It's unused, unmaintained, and confuses people.

Closes #97


egekorkan:
I agree with making this obsolete. Even though nothing is actually deleted in git, I would prefer to move this into an archive folder since some work is done before and would be good to have this as a document to show as an answer to "Why haven't you done it like HTTP in RDF?"

JKRhb:
Just to clarify: Does this PR imply that there should be no CoAP ontology at all (since it closes #97)? Or would we simply start from scratch for eventually creating such a document?

ektrah:
@egekorkan Moving it to an archive folder works for me. Is there already a place for this or where should this go?

@JKRhb For the purposes of the CoAP Binding, there should be no "CoAP in RDF" ontology at all IMHO. This doesn't preclude a "CoAP in RDF" ontology being created in a different context in the future, which could be based on the work done previously here or start from scratch.

relu91:
Removing without a valid replacement is problematic. Where will the terms used in the CoAP binding template document be defined? We need at least a `context` file.

egekorkan:
Call of 16.11: We can move it to the archive to not delete and provide an ontology later. It can be simple one to one mapping instead of a real ontology but at the same time, our CoAP ontology should be significantly smaller than HTTP in RDF so not too much work is needed.

egekorkan:
@ektrah you can use https://github.com/w3c/wot-binding-templates/tree/main/Archive for the archive

ektrah:
Superseded by #204

PR #190, opened 2022-10-12
Align protocol binding abstracts

egekorkan:
This PR touches all protocol bindings, thus I have requested review from all current editors and active contributors. However, there is no obligation to read and review all of them. It would be best if you can review the documents you have contributed to and the template, that would be enough.

My reasoning is to use a single template that changes per protocol binding. Having a common image is important in my opinion since all these documents can be accessed independent of the other and of the main document.

### DIFF
--- a/bindings/protocols/mqtt/index.html
+++ b/bindings/protocols/mqtt/index.html
@@ -63,7 +63,7 @@
         <p>
             In the context of the Web of Things (WoT), a Binding Template is a blueprint that gives guidance on how to 
             implement a specific IoT protocol, data format or IoT platform. 
-            This document gives implementation guidelines regarding the MQ Telemetry Transport (MQTT) protocol, which is
+            This document gives implementation guidelines regarding the MQTT protocol, which is
             a lightweight, asynchronous, transport-independent, open protocol for publishing and subscribing to messages. 
             It has been employed in many IoT platforms and IoT applications for its simplicity and scalability.
         </p>
--- a/bindings/protocols/mqtt/index.html
+++ b/bindings/protocols/mqtt/index.html
@@ -60,11 +60,18 @@
 
 <body>
     <section id="abstract">
-      In the context of the Web of Things, a Binding Template is 
-      a blueprint that gives guidance on how to implement a specific IoT protocol, data format or IoT platform. 
-      This document gives implementation guidelines regarding the MQTT protocol. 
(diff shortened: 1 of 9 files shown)


bindings/protocols/mqtt/index.html
```diff
@@ -60,11 +60,18 @@
...
     <section id="abstract">
-      In the context of the Web of Things, a Binding Template is 
-      a blueprint that gives guidance on how to implement a specific IoT protocol, data format or IoT platform. 
-      This document gives implementation guidelines regarding the MQTT protocol. 
-      The MQTT protocol is a lightweight, asynchronous, transport-independent, open protocol for publishing and subscribing to messages. 
-      It has been employed in many IoT platforms and IoT applications for its simplicity and scalability.
+        <p>
+            In the context of the Web of Things (WoT), a Binding Template is a blueprint that gives guidance on how to 
+            implement a specific IoT protocol, data format or IoT platform. 
+            This document gives implementation guidelines regarding the MQ Telemetry Transport (MQTT) protocol, which is
```
JKRhb, line 66: If I am not mistaken, MQTT nowadays is not an abbreviation anymore. See page 2 of [these OASIS minutes](https://www.oasis-open.org/committees/download.php/49028/OASIS_MQTT_TC_minutes_25042013.pdf):

> The TC agrees that MQTT should not be standing for anything.

```suggestion
            This document gives implementation guidelines regarding the MQTT protocol, which is
```

bindings/protocols/mqtt/index.template.html
```diff
@@ -60,11 +60,18 @@
...
     <section id="abstract">
-      In the context of the Web of Things, a Binding Template is 
-      a blueprint that gives guidance on how to implement a specific IoT protocol, data format or IoT platform. 
-      This document gives implementation guidelines regarding the MQTT protocol. 
-      The MQTT protocol is a lightweight, asynchronous, transport-independent, open protocol for publishing and subscribing to messages. 
-      It has been employed in many IoT platforms and IoT applications for its simplicity and scalability.
+        <p>
+            In the context of the Web of Things (WoT), a Binding Template is a blueprint that gives guidance on how to 
+            implement a specific IoT protocol, data format or IoT platform. 
+            This document gives implementation guidelines regarding the MQ Telemetry Transport (MQTT) protocol, which is
```
JKRhb, line 66: (See above.)

```suggestion
            This document gives implementation guidelines regarding the MQTT protocol, which is
```

JKRhb (commented):
Looks good to me, see three minor comments below :)

PR #193, opened 2022-10-18
Alternative proposal for handling CoAP Content-Formats

JKRhb:
This PR makes another proposal for handling CoAP Content-Formats in Thing Descriptions. In order to simplify the vocabulary and delegate the exact semantics to the main TD document, I only defined one extra vocabulary term called `cov:contentFormat` which can be used as an alternative to `contentType` and `contentCoding`.

The exact algorithm for when to use the Accept option is intentionally omitted, as it should follow the same logic as for the corresponding HTTP header. However, an example is given how the Content-Format and the Accept option would be used by Consumers in practice.

I think this could be a simpler solution than #188 as it builds upon existing features, adding only an additional layer for a more accurate mapping. I am looking forward to your feedback and further discussions, though :)

### DIFF
--- a/bindings/protocols/coap/index.html
+++ b/bindings/protocols/coap/index.html
@@ -128,11 +128,16 @@
 </pre>
             <p>
                 The <code>contentType</code> and <code>contentCoding</code> members of a form specify the CoAP content format of the request and/or response.
-                Consumers are expected to be able to map the combinations of content type and content coding they understand to the corresponding CoAP content format IDs listed in the <a href="https://www.iana.org/assignments/core-parameters/#content-formats">IANA CoAP Content-Formats registry</a>.
-                Producers are advised to only use combinations of content types and content codings in CoAP forms that have such a content format ID available.
+                In <abbr title="Constrained Application Protocol">CoAP</abbr>, combinations of content type and content coding are encoded using standardized values listed in the <a href="https://www.iana.org/assignments/core-parameters/#content-formats">IANA CoAP Content-Formats registry</a>.
+                Therefore, not every combination of content type and content coding can be used with CoAP.
+                Furthermore, even parameters are included in the mapping: in [[[#example-request-defaults]]], the content type <code>text/plain;charset=utf-8</code> would be mapped to Content-Format ID 0.
+                Omitting the <code>charset</code> parameter, for instance, would break a literal mapping from content type to Content-Format in this case.
+            </p>
+            <p>
+                Section [[[#content-negotiation]]] introduces the additional vocabulary terms <code>cov:contentFormat</code> and <code>cov:accept</code> to deal with ambiguities of content types and content codings in the context of CoAP, and to give Consumers guidance for the content negotiation process.
             </p>
         </section>
-        <section>
+        <section id="content-negotiation">
             <h3>Content Negotiation</h3>
             <p>
                 Content negotiation in CoAP is used to negotiate the representation of CoAP resources that may have different representations available.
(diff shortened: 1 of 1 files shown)


ektrah:
Could you show an example of (1.) a property where the client can choose between 3 different formats when reading it and (2.) a property where the client can choose between 3 different formats when writing it?

ektrah:
(This PR relates to the discussion in https://github.com/w3c/wot-thing-description/pull/1564.)

JKRhb:
> Could you show an example of (1.) a property where the client can choose between 3 different formats when reading it and (2.) a property where the client can choose between 3 different formats when writing it?

I am afraid this is a bit verbose, but I assume something that could look something like this:

```json
{
    "@context": "https://www.w3.org/2022/wot/td/v1.1",
    "title": "Test",
    ...
    "properties": {
        "property1": {
            "forms": [
                {
                    "href": "coap://example.org/form1",
                    "op": "readproperty",
                    "contentType": "application/json",
                    "cov:contentFormat": 50
                },
                {
                    "href": "coap://example.org/form2",
                    "op": "readproperty",
                    "contentType": "application/cbor",
                    "cov:contentFormat": 60
                },
                {
                    "href": "coap://example.org/form3",
                    "op": "readproperty",
                    "contentType": "application/xml",
                    "cov:contentFormat": 41
                }
            ]
        },
        "property2": {
            "forms": [
                {
                    "href": "coap://example.org/write-form1",
                    "op": "writeproperty",
                    "contentType": "application/json",
                    "cov:contentFormat": 50
                },
                {
                    "href": "coap://example.org/write-form2",
                    "op": "writeproperty",
                    "contentType": "application/cbor",
                    "cov:contentFormat": 60
                },
                {
                    "href": "coap://example.org/write-form3",
                    "op": "writeproperty",
                    "contentType": "application/xml",
                    "cov:contentFormat": 41
                }
            ]
        }
    }
}
```

relu91:
> In the case that no explicit <code>cov:contentFormat</code> is defined (see section [[[#content-negotiation]]]), Consumers are expected to be able to map the combinations of content type and content coding they understand to the corresponding CoAP content format IDs listed in the <a href="https://www.iana.org/assignments/core-parameters/#content-formats">IANA CoAP Content-Formats registry</a>.

Just to understand, given that sentence, this means that in the trivial case ( dev forgets/don't want to define `cov:contentFormat`) the runtime should contain the logic to map correctly the content type string to CoAP's IDs, right? So what is the benefit of explicitly state `"cov:contentFormat"`. Also, should we somehow enforce that `cov:contentFormat` should be filled accordingly to `contentType`?

JKRhb:
Thanks, @relu91, that are very valid questions. I think including the numeric Content-Format could make it easier for consumers, e.g., to set the Accept option accordingly, even if they don't have a full registry for mapping the string representations to the numeric ones. Also, this could reduce potential ambiguity when mapping from string values to numeric codes (since I got the impression that this mapping process is not that well-defined yet). I was also thinking about removing the sentence regarding consumer expectations or limiting it to a certain type of consumer (e.g., one that does not have resource constraints).

> Also, should we somehow enforce that cov:contentFormat should be filled accordingly to contentType?

That is a very good suggestion, could we add a `MUST` assertion for this requirement (if this is possible in informative documents)?

bindings/protocols/coap/index.html
```diff
@@ -142,23 +142,35 @@ <h3>Content Negotiation</h3>
                 The CoAP Accept option is used by clients to request a particular content format, while the <code>Content-Format</code> option is used by clients and servers to indicate the content format of the representation in requests and responses, respectively.
             </p>
             <p>
-                The ... can be used to provide a list of one or more content formats that the Thing is able to provide in a CoAP response for the operation.
-                The Consumer selects the most suitable one and sets the <code>Accept</code> option in the request accordingly.
+                Since mapping the string-based <code>contentType</code> and <code>contentCoding</code> fields to a numeric Content-Format can be prone to error, this document defines a vocabulary for directly defining the Content-Format(s) associated with an affordance.
+                To do so, the term <code>cov:contentFormat</code> can be used in the same places as <code>contentType</code> and <code>contentCoding</code>.
```
egekorkan, line 146: `contentType` must be present otherwise the consumer must assume json and this will cause problems

egekorkan:
We should not use `<em class="rfc2119" "MUST">MUST</em>` but we can define our own CSS class or use lowercase if we really need it.

@relu91 said: if we introduce this new keyword, validation might be harder since you still need a copy of the registry to match content-format and content type.

@egekorkan : If the consumer trusts the keywords it knows, it is fine.

@ektrah could you have another look at this after the changes?

egekorkan (approved):
To me this looks good as spoken last week, I would prefer to have Klaus' feedback before proceeding.

egekorkan:
Call of 23.11:
- The proposal looks good with the changes discussed:
  - An example like at [here](http://plugfest.thingweb.io/playground/#C4EwVgzg9gdg3gKAAQqQIgAIGNbAKYAewaAXOgBbDAAOEJA9PQO4sB0TAzK1AE4Dm9AEwAGQYOZRg9UPQBuARlby0AGmSo0AQyzAAlrAikkiVKfQRgmnsTImz9tLpjUArjePpgAT2p4jjmHw+PB40JABfNXsHKDdXdzhPHz8yNAseJz4wyPVojQAzXgBbQzIAbVy8szsqqrRyHjx8-xxNagZ6MpFheRIAEQAhAA4SEnkAXXoLK2Io2rqcWRIcQLxAgDFizXcAVmE5+ei0RZJtLDxqGwA2fcrD0zRGiGoDFOM7+4cV-ECAFWT-G1qAAbXRYbb6GD0LAAI14qg+nweJ2+a2Amx4RW2RhuiPu4TxeRySJQNRJFEazVSrXajC6wh6-WGowmU0s1gR5I0KNwaIxWN2ty5KGOUCWZwuNj2B3Jjzwz1eRjJwvQqL+ANSQNB4L0sHokFgnJV3LFy15Gy2gsJ8wJJNt83GePtZkqBPCQA) should be used to explain why the accept header would be needed. 
  - Accept header term is also needed to be introduced (see #188).
  - If there is accept header, Consumer must use it.
- Similar should be done for HTTP Binding as well

egekorkan (changes requested):
Also, `cov:accept` is missing from the table at Chapter 4.1

egekorkan:
@JKRhb could you use different commit messages? It might also be some error somewhere since the messages are exactly the same.

JKRhb:
@egekorkan Oh, sure, with the fixup commits I actually intend to squash them before merging (using `git rebase -i --autosquash main`) with a slightly adjusted commit message. Would you prefer keeping the recent changes as separate commits?

egekorkan:
> @egekorkan Oh, sure, with the fixup commits I actually intend to squash them before merging (using `git rebase -i --autosquash main`) with a slightly adjusted commit message. Would you prefer keeping the recent changes as separate commits?

ok I did not know, it is fine :)

JKRhb:
@egekorkan @ektrah I updated the PR – let me know if there is anything that should be changed :)

bindings/protocols/coap/index.html
```diff
@@ -142,23 +147,57 @@ <h3>Content Negotiation</h3>
...
+                If the field <code>cov:contentFormat</code> is present, it must match the string representation expressed by <code>contentType</code> and <code>contentCoding</code>.
+                Corresponding with the default <code>contentType</code> value <code>application/json</code>, Consumers are supposed to assume a Content-Format value 50 if <code>cov:contentFormat</code> and <code>application/json</code> is not present.
             </p>
-            <pre id="example-request-format" class="example" title="...">
-{
-    ...
-}
-</pre>
+            <p>
+                Consumers may use the <code>cov:contentFormat</code> in a Form's <code>response</code> member object to derive the value for an Accept option in their request.
```
ektrah, line 161: IIRC we decided that the value for an Accept option would not be derived from anything but always needs to be given explicitly, no?

ektrah (commented):
I have one question and a bunch of editorial nits. For the editorial nits I think it would be easiest if I simply open a pull request after this one is merged. For the question, see below.

bindings/protocols/coap/index.html
egekorkan: I think that our decision was that if it is given explicitly in the TD, it must be used by the Consumer. If the Thing is not handing the accept header, I do not see a point why it must be included in its TD.

bindings/protocols/coap/index.html
```diff
@@ -142,23 +147,57 @@ <h3>Content Negotiation</h3>
...
+                However, TDs may also contain a <code>cov:accept</code> to explicitly require a Consumer to provide an Accept option.
```
JKRhb, line 162: @ektrah @egekorkan I guess we can simply remove the sentence:

```suggestion
                Forms may also contain a <code>cov:accept</code> field to explicitly require a Consumer to provide an Accept option.
```

However, I think one concern I still had was that the `cov:accept` field leads to some redundancy, since its value should probably always be derivable from either the form or the response member. Therefore, I am wondering if this field could also be a boolean, which would indicate that an Accept option must be used.

Then again, it is not that big of a deal, and we could also just this version in order to not prolong the discussion more than needed.

ektrah (approved):

egekorkan:
Call of 07.12:
- @ektrah has approved the changes, we can merge.

PR #204, opened 2022-11-23
Move Coap Ontology

egekorkan:
Takesover #189 and just moves it to the archive folder


ektrah (approved):

egekorkan:
Call of 23.11: 
- Agreed to merge

PR #210, opened 2022-11-29
Create CODEOWNERS

ektrah:
Code owners are automatically requested for review when someone opens a pull request that modifies code that they own.

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,4 @@
+/bindings/protocols/coap/ @ektrah
+/bindings/protocols/http/ @egekorkan
+/bindings/protocols/modbus/ @relu91
+/bindings/protocols/mqtt/ @egekorkan @relu91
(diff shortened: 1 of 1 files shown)


egekorkan:
I definitely like the idea and we should do this in all WoT repo. Two points:
- There is an error displayed but I do not totally understand the real meaning
- I think that the codeowners should be strictly correlated with the list of editors of each binding and we should document this in the readme as well.

relu91:
> There is an error displayed but I do not totally understand the real meaning

From the doc: 
>The people you choose as code owners must have write permissions for the repository

So basically the error is because the cited people don't have write permissions. Not sure if it is just a warning or if it completely disables the functionality for such users. 

Other than this I really like this feature! 👍🏻

Issue #220, opened 2022-12-07
Bringing back temporary sections into the main body


egekorkan:
We should bring the https://w3c.github.io/wot-binding-templates/#sec-temp back into the main part of the document. They contain guidelines and explanations on how the binding templates work in general.

egekorkan:
These are in the core document now, closing

PR #223, opened 2022-12-21
Orphaned Section Reorg - Part 1: Protocols

egekorkan:
I have started working on #220. Since there are many sections there, in order to allow reviews, I want to split it as much as possible. In this PR, I am addressing section B.1 at https://w3c.github.io/wot-binding-templates/#form-element . 

To make it easier to review, I have made sure to split commits into meaningful sections. So for an in-depth review, please look at changes per commit.

Working on this PR made me question the need for [B.3 Interaction Affordances](https://w3c.github.io/wot-binding-templates/#sec-interaction-patterns) which is what this section does but per affordance type. I would like to get feedback if we need that.

Subprotocol orphaned section is still there. I am guessing that it will go into the protocol bindings section as well.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-binding-templates/pull/223.html" title="Last updated on Dec 21, 2022, 5:02 PM UTC (dabd904)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-binding-templates/223/a3e0636...dabd904.html" title="Last updated on Dec 21, 2022, 5:02 PM UTC (dabd904)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -941,8 +941,8 @@
         
     </section>
 
-    <section id="sec-orphaned">
-        <h2>Orphaned Sections</h2>
+    <section id="sec-temp">
+        <h2>Temporary Sections</h2>
 
         <p class="ednote">
             The below chapters are copied from other places in this document. They require discussion on where to place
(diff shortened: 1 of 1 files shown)


egekorkan:
In the meantime, it would be better to not use "orphaned section". We can use "temporary" for now.

PR #225, opened 2023-01-25
JSON Schema usage in Bindings

### LINKED ISSUE #205: JSON Schemas of Protocol Binding Templates
Currently, we have [a JSON Schema](https://github.com/w3c/wot-binding-templates/blob/main/bindings/protocols/modbus/modbus.schema.json) for the Modbus protocol binding template. The other protocols do not have it at the moment.
The schema at Modbus does validation on top of the TD schema by making sure that all the forms conform to the Modbus schema.

Overall, we should discuss how to do this kind of validation. The current way in the Modbus schema would give fail if there is an HTTP and Modbus form inside one TD. So in general, we can validate a form only if the URI scheme matches and then the vocabulary terms on top. Also, if we agree on this, we should put a recommendation for all binding templates to provide a JSON Schema for that specific form.

egekorkan:
fixes #205 

Note: Each commit can be viewed independently for all the different implemented features. 

In this PR, I have done the following:
- Add a requirement that a JSON Schema should be provided for the protocol bindings
- Create a JSON Schema template
- Add link to Modbus schema in the modbus document (@relu91)
- Fixed a bug in the render script that was asking for coap.ttl

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-binding-templates/pull/225.html" title="Last updated on Jan 25, 2023, 11:45 AM UTC (7fc8fbf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-binding-templates/225/a3e0636...7fc8fbf.html" title="Last updated on Jan 25, 2023, 11:45 AM UTC (7fc8fbf)">Diff</a>

### DIFF
--- a/bindings/index.template.html
+++ b/bindings/index.template.html
@@ -24,7 +24,16 @@
                     href: "../../../index.html"
                     }
                 ]
-                }
+                },
+                {
+                  key: "JSON Schema",
+                  data: [
+                      {
+                      value: "BINDINGNAME JSON Schema",
+                      href: "template.schema.json"
+                      }
+                  ]
+              }
             ]
         };
     </script>
--- a/bindings/protocols/modbus/index.html
+++ b/bindings/protocols/modbus/index.html
@@ -36,6 +36,15 @@
                       href: "../../../ontology/modbus.html"
                       }
                   ]
+              },
+              {
+                  key: "JSON Schema",
+                  data: [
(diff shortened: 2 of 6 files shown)


egekorkan:
Call of 25.01: @relu91 and @mjkoster approve it, we can merge.

PR #228, opened 2023-02-01
Temporary Section Reorg - Part 2: Payloads

egekorkan:
As sort of a follow up of #223, I am doing the same for the payload bindings. I am starting with the content types and adding some new examples as well.

### DIFF
--- a/index.html
+++ b/index.html
@@ -160,10 +160,10 @@
              Description[[WOT-THING-DESCRIPTION]] (TD), with Things that exhibit diverse protocols, payload formats and their usage.
             These subspecifications are categorized into three:
         <ul>
-            <li> <a href="#protocol-intro"></a>: Application layer protocols (e.g., HTTP[[?RFC7231]], CoAP[[?RFC7252]],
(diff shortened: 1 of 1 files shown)


egekorkan:
Call of 01.02:
- Since the forms are repeated except for the `contentType`, the examples become too verbose. We can change it to a bullet list with the different options. A single JSON example at the end can be left.

egekorkan:
@danielpeintner and @relu91 I have reorganized it as discussed. Please have a look.

index.html
```diff
@@ -420,71 +422,156 @@ <h4>Creating a new Protocol Binding Template Subspecification</h4>
...
+                            "forms":[
+                            {
+                                "href": "http://example.com/properties/temperature",
+                                "op": "readproperty",
+                                "contentType": "application/json"
+                            },
+                            {
+                                "href": "coap://example.com/properties/temperature",
+                                "op": "readproperty",
+                                "contentType": "application/cbor;charset=utf-8"
```
danielpeintner, line 487: I am not a CBOR expert but i was surprised to read 
`contentType": "application/cbor;charset=utf-8`
since CBOR is binary-encoded. Isn't that the case?
Maybe `charset` is better suited for `text/plain;charset="UTF-8`

egekorkan: I have fixed this

Issue #207, opened 2022-11-29
MQTT verbs with or without prefixes


egekorkan:
Currently, the MQTT binding vocabulary uses `"mqv:controlPacket": "mqv:subscribe"` and thus prefixes both terms. However, this is not how it is done in HTTP, CoAP and Modbus. @relu91 what was the initial reason to do so? Can we align them all to use the prefix only in the key and not in the value?

PR #234, opened 2023-02-06
Fix MQTT issues

### LINKED ISSUE #230: JSON Schema for MQTT Vocabulary
As a follow up of https://github.com/w3c/wot-binding-templates/pull/225, we need one for MQTT.

relu91:
This PR fixes #230 and #207. I decided to tackle the two issues together because they were pretty interrelated and it was easier to resolve both at once. 

**Issues**: It seems that the Modbus file had wrong line endings, I had to put this change in a specific commit to avoid bringing it over on some other unrelated changes. I also added a small comment on the readme referencing the new schema template.


danielpeintner (commented):
I think the changes for `ontology/modbus.html` in this MQTT PR were not intentional.

relu91:
> I think the changes for ontology/modbus.html in this MQTT PR were not intentional.

Yes, they weren't but check the previous comment:
> Issues: It seems that the Modbus file had wrong line endings, I had to put this change in a specific commit to avoid bringing it over on some other unrelated changes.

I didn't have time to properly check why the line endings end up being wrong, but if it is a blocker I will investigate more.

egekorkan (approved):

PR #237, opened 2023-02-08
Add JSON Schema for HTTP

### LINKED ISSUE #229: JSON Schema for HTTP Binding Terms
As a follow up of https://github.com/w3c/wot-binding-templates/pull/225, we need one for HTTP.

egekorkan:
fixes #229 

There is one annoying issue here though. The HTTP context is actually embedded in the TD one. So if we make it mandatory to have the `"htv":"http://www.w3.org/2011/http#"` TDs that do not put it but still use the HTTP terms will be invalid but they are actually ok. Any opinions?

### DIFF
--- a/bindings/protocols/http/http.schema.json
+++ b/bindings/protocols/http/http.schema.json
@@ -52,6 +52,9 @@
         },
         "events": {
             "$ref": "#/definitions/affordance"
+        },
+        "@context": {
+            "$comment":"When the HTTP vocabulary is not part of the TD vocabulary anymore, requirements on the context usage should be added"
         }
     }
 }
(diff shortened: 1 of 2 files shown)


JKRhb:
> There is one annoying issue here though. The HTTP context is actually embedded in the TD one. So if we make it mandatory to have the `"htv":"http://www.w3.org/2011/http#"` TDs that do not put it but still use the HTTP terms will be invalid but they are actually ok. Any opinions?

Would including the TD context URL as an alternative be an option here?

egekorkan:
Call of 08.02:
- We will allow the simple TD context. Thus, simply removing the value requirement for the context

egekorkan:
Call of 22.02:
- We should add a comment to say that we should bring back the context check.
- @egekorkan will add the comment and merge

Issue #216, opened 2022-11-30
Instructions on how to create a vocabulary and publish should be provided


egekorkan:
Instructions on how to create a vocabulary and publish should be provided as well in another PR

_Originally posted by @egekorkan in https://github.com/w3c/wot-binding-templates/issues/212#issuecomment-1332422933_

JKRhb:
Would it make sense to add the instructions to the `CONTRIBUTING.md` file? Or is there a better place for them (e.g., a wiki created for the repository?)

egekorkan:
> Would it make sense to add the instructions to the `CONTRIBUTING.md` file? Or is there a better place for them (e.g., a wiki created for the repository?)

So both of them would not be desirable:
- Contributing.md might be fine but it is more about the license-like stuff
- wiki is generally not fine because it is not reviewable with PRs etc.

I think that a separate md file in the root of the repository should be good.

egekorkan:
Call of 05.04:
- Given that other entities can create it and host themselves, it is not contributing to this repo but that should be linked to CONTRIBUTING.md
- We should look into the other documents registry section instructions to avoid conflicts in the future.

PR #240, opened 2023-02-08
feat(coap): add CoAP JSON-LD @context

JKRhb:
This PR contains a first attempt at creating a JSON-LD @context file for the CoAP Vocabulary. It is honestly mostly the result of some trial-and-error using the JSON-LD playground, so I am not really sure if the @context is usable already 😅 I was also a bit unsure whether the @context for a Binding Document is a result or an input of the rendering process. So these points need to be taken into account when reviewing the PR.

I hope the file can still serve as a starting point for further discussion and for working out a @context file that is actually usable :) I am looking forward to your feedback!

### DIFF
--- a/bindings/protocols/coap/context.jsonld
+++ b/bindings/protocols/coap/context.jsonld
@@ -1,0 +1,43 @@
+{
+    "@context": {
+        "xsd": "http://www.w3.org/2001/XMLSchema#",
+        "cov": "https://www.w3.org/2019/wot/cov#",
+        "method": {
+            "@id": "cov:hasMethod",
+            "@type": "xsd:string"
+        },
+        "BlockWiseTransferParameters": {
+            "@id": "cov:BlockWiseTransferParameters"
+        },
+        "blockwise": {
+            "@id": "cov:hasBlockwise",
+            "@type": "@id",
+            "@container": "@index"
+        },
+        "qblockwise": {
+            "@id": "cov:hasQBlockwise",
+            "@type": "@id",
+            "@container": "@index"
+        },
+        "block2SZX": {
+            "@id": "cov:hasblock2SZX",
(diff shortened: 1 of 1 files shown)


egekorkan:
I think this is not the correct direction and I am sorry that you had no chance to know about that since https://github.com/w3c/wot-binding-templates/issues/216 is still open. The goal is to auto generate the ontology by using coap.ttl mapping.ttl and templates.sparql files. You can find how they should look like in Modbus:
- https://github.com/w3c/wot-binding-templates/blob/main/bindings/protocols/modbus/mapping.ttl 
- https://github.com/w3c/wot-binding-templates/blob/main/bindings/protocols/modbus/template.sparql
- https://github.com/w3c/wot-binding-templates/blob/main/ontology/modbus.ttl

Note: Also see what the render.js file requires: https://github.com/w3c/wot-binding-templates/blob/main/render.js

@relu91 please confirm about the direction as well

JKRhb:
Thank you for the feedback, @egekorkan! I was a bit unsure how the process is supposed to work, since `render.js` complained for some reason that `context.jsonld` was missing. That gave me the impression that – unintuitively – the JSON-LD context would be missing as an input. I guess some refactoring of the generation script could also make things a bit easier to understand.

I will close the PR for now and try again from the other direction :)

relu91:
Actually, this was correct, the render script does not autogenerate the context file, it is a manual operation that needs to be done as explained in https://github.com/w3c/wot-binding-templates/blob/main/bindings/README.md

egekorkan:
@JKRhb did you really want to close this PR by deleting the branch?

JKRhb:
> @JKRhb did you really want to close this PR by deleting the branch?

Sorry, I had some technical difficulties, I will open a new PR ASAP :)

PR #241, opened 2023-02-09
Refactor render script

JKRhb:
As mentioned in https://github.com/w3c/wot-binding-templates/pull/240#issuecomment-1423365415, this PR makes a proposal for refactoring the render script using async-await and object destructuring, which should make the code a bit more readable.

I could not get rid of the promiseChain, which I also found a bit confusing to read, without breaking the script, but I still think it would also be a good candidate for future refactoring.

### DIFF
--- a/render.js
+++ b/render.js
@@ -1,86 +1,104 @@
-let fs = require('fs');
-let sttl = require('sttl');
-let urdf = require('urdf');
-const toRDf = require('./toRdf.js').context
+const fs = require('fs');
+const sttl = require('sttl');
+const urdf = require('urdf');
+const toRDf = require('./toRdf.js').context;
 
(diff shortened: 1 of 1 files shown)


relu91 (approved):
Looks good. There is still room for improvement but it has already better readability.

egekorkan (approved):

egekorkan:
@JKRhb did you really want to close this PR by deleting the branch?

egekorkan:
@JKRhb should we merge this or #241 first? Is the code exactly same?

JKRhb:
> @JKRhb should we merge this or #241 first? Is the code exactly same?

The changes are also included in #246, so either PR could be merged to bring the commit with the changes to the main branch :) However, this PR should probably be merged first to achieve a cleaner diff in #246 -- I would rebase its branch afterwards.

PR #242, opened 2023-02-09
Generate CoAP vocabulary from RDF

JKRhb:
Based on the feedback by @egekorkan in https://github.com/w3c/wot-binding-templates/pull/240 I started working on a new, simplified CoAP ontology for generating the Binding Template and (potentially) the JSON-LD context automatically. This PR is still in the early stages of development, but I wanted to open it already as a draft to show that there is something in progress. (Since the @context needs to be created manually at the moment, it also contains the JSON-LD file from the last PR.)

Perhaps we could modify the rendering script to automatically generate JSON-LD `@context`? I think this would greatly simplify the process of adding a new binding document.

(This PR depends on #241 and needs to be rebased once that PR is merged.)


JKRhb:
After a lot of fiddling around, I think this PR would now be ready for an initial review. There are still a number of issues that need to be addressed (e.g., the enum values of the blockwise parameters) but the render script should already be able to recreate the current state of the document.

One thing that certainly needs to be discussed is the scope of the CoAP Ontology and whether it should actually be called "CoAP in RDF", since it probably will only cover a subset of CoAP.

JKRhb:
> There are still a number of issues that need to be addressed (e.g., the enum values of the blockwise parameters) but the render script should already be able to recreate the current state of the document.

I now managed to also extract the blockwise enum values from the Ontology (probably not in the most elegant way, but at least it works for now 😅). I would assume that it should mostly be refinement and refactoring that is left now :)

Issue #217, opened 2022-11-30
@context for CoAP vocabulary 


ektrah:
The CoAP protocol binding template doesn’t come with a “CoAP in RDF” ontology. Nevertheless, there need to be the necessary `@context` definitions so that the vocabulary can be used in JSON-LD.

ektrah:
As a first step, #246 provides a [context.jsonld](https://github.com/w3c/wot-binding-templates/blob/c314cbd4e8c7606e7461bd5ecec64f4f16b6dc7f/bindings]/protocols/coap/context.jsonld) file. Next, the CoAP binding document should be updated to describe how to use this file in a TD. There may be other adjustments needed to make it work; see https://github.com/w3c/wot-binding-templates/pull/246#pullrequestreview-1333140744. A good test could be if you can paste a TD using the CoAP vocabulary into the JSON-LD Playground and all JSON keys are resolved properly.

relu91:
On the TD repo, I did some unit tests to check jsonld correctness, we could introduce something similar also for each binding.

PR #246, opened 2023-02-15
Generate CoAP vocabulary from RDF

JKRhb:
Building upon #241, this PR makes a first attempt at providing a CoAP JSON-LD @context and generating the vocabulary terms from an ontology.

While the render script is already capable of reproducing the current state of the CoAP Binding Template using the Ontology, the `mapping.ttl` and the `template.sparql` files, there is definitely still room for improvement. Furthermore, the scope of the CoAP ontology has to be (re)defined. Still, this PR can probably be used as a basis for future work in this regard.


relu91:
Can you explain more about how the `@context` generation works?

JKRhb:
> Can you explain more about how the `@context` generation works?

Oh, the `@context` is currently the same as in #240, so there is no automatic generation process involved (yet). Based on your comment in https://github.com/w3c/wot-binding-templates/pull/240#issuecomment-1423823672, I thought that automation could be part of a follow-up PR if that would be okay :)

relu91:
Ah ok, I was confused by the title 😃

JKRhb:
Oh, I see, sorry 😅 I slightly adjusted the PR description, I hope it is now a bit clearer what is manual and what is automatic :)

ontology/coap.html
```diff
@@ -0,0 +1,52 @@
+
+<!DOCTYPE html>
+<html>
+    <head>
+    <meta charset='utf-8'>
+    <script src='https://www.w3.org/Tools/respec/respec-w3c-common' class='remove'></script>
+    <script class='remove'>
+        var respecConfig = {
+            specStatus: "ED",
+            editors: [{name:"Victor Charpenay",company:"Siemens AG",companyURL:"https://www.siemens.com/"},{name:"Ege Korkan",company:"Siemens AG",companyURL:"https://www.siemens.com/"},{name:"Sebastian Käbisch",company:"Siemens AG",companyURL:"https://www.siemens.com/"},{name:"Michael Koster",company:"Invited Expert",companyURL:"https://github.com/mjkoster/"},{name:"Jan Romann",company:"Invited Expert",companyURL:"https://github.com/JKRhb/"}],
```
egekorkan, line 10: I am not sure that we should use this list of authors. I think that the editors of the CoAP binding should be the authors too.

egekorkan (changes requested):
Also, please link the document ontology to the core document.

Another nice observation is that there is actually nearly no diff on the coap/index.html :)

ontology/coap.html
JKRhb: Very good point, for now I simply copied over the old author and contributer list. Should the currently listed contributors be kept? Or should we start from scratch here?

egekorkan: I would say from scratch. You should add your name though. I have talked with @ektrah and he is ok to make you another author of the coap binding but we will discuss this separately at https://www.w3.org/WoT/IG/wiki/WG_WoT_Thing_Description_WebConf#February_22.2C_2023

egekorkan:
also this fixes #217

JKRhb:
> Also, please link the document ontology to the core document.

Done in 8da157b5dbd57ce9e70ceee9240da5b8c2416c59 :)

> also this fixes #217

I edited the PR description to close the issue automatically.

ontology/coap.html
JKRhb: Thank you for the feedback! I started to adjust the author lists in 1a82bc5bcb75a7eb6204b91036edf24194c0ed4f

egekorkan:
Call of 22.02:
- We will wait for @ektrah to review but it can be merged without a meeting
- The group is ok to make @JKRhb another editor of the CoAP binding.

ontology/coap.ttl
```diff
@@ -0,0 +1,165 @@
...
+                                                      schema:affiliation [ rdf:type schema:Organization ;
+                                                      schema:name "Invited Expert" ;
+                                                      schema:url "https://github.com/JKRhb/"
+                                                    ] ;
+                                                 schema:name "Jan Romann"
+                                               ] ;
+                                    vann:preferredNamespacePrefix "cov" ;
+                                    vann:preferredNamespaceUri : ;
+                                    rdfs:comment "Vocabulary to represent CoAP messages in RDF."@en ;
+                                    rdfs:label "CoAP in RDF"@en .
```
ektrah, line 27: The [HTTP Vocabulary in RDF 1.0](https://w3c.github.io/wot-binding-templates/bindings/protocols/http/index.html) describes itself as follows (emphasize mine):

> This document provides a representation of the HTTP vocabulary in the Resource Description Framework (RDF), to allow quality assurance tools to **record the HTTP headers that have been exchanged between a client and a server**.

In other words, it provides a way to convert exchanged HTTP messages into RDF graphs. With a bit of imagination, this can also be (ab)used to serialize HTTP messages that haven't yet been exchanged, but the focus is clearly on serializing HTTP headers.

In the case of the vocabulary in the CoAP Binding document, the vocabulary is *not* intended for serializing exchanged CoAP options into RDF. The vocabulary instead is used to describe a CoAP server's capabilities and expected parameters for a WoT operation.

I therefore feel it is inappropriate to call this vocabulary "CoAP in RDF".

ontology/coap.ttl
```diff
@@ -0,0 +1,165 @@
...
+#################################################################
+
+:BlockWiseTransferParameters rdf:type owl:DatatypeProperty ;
+        rdfs:domain :Message ;
+        rdfs:isDefinedBy <> ;
+        rdfs:label "BlockWiseTransferParameters"@en .
+
+:hasblock2SZX rdf:type owl:DatatypeProperty ;
+        rdfs:domain :BlockWiseTransferParameters ;
+        rdfs:range :Blocksize ;
```
ektrah, line 87: The SHACL shape in [mapping.ttl](https://github.com/w3c/wot-binding-templates/pull/246/files#diff-bf9dcd7ed26a879e846cf4b2fbf9496df8e6585b146d5f3c3e4738e8926be483R80) requires that the value has datatype `xsd:unsignedShort`. So it seems either the SHACL shape is not satisfied when the value is an instance of `:Blocksize` or the OWL reasoner would try to make a block size like `"16"^^xsd:unsignedShort` an instance of `:Blocksize`, right?

bindings/protocols/coap/mapping.ttl
```diff
@@ -0,0 +1,87 @@
...
+@prefix xml: <http://www.w3.org/XML/1998/namespace> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+@prefix foaf: <http://xmlns.com/foaf/0.1/> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
+@prefix vann: <http://purl.org/vocab/vann/> .
+@prefix dcterms: <http://purl.org/dc/terms/> .
+@prefix schema: <http://schema.org/> .
+@prefix sh: <http://www.w3.org/ns/shacl#> .
+@prefix coap: <https://www.w3.org/2019/wot/coap#> .
```
ektrah, line 18: Should the prefix be `cov:` for consistency?

```suggestion
@prefix cov: <https://www.w3.org/2019/wot/coap#> .
```

bindings/protocols/coap/template.sparql
```diff
@@ -0,0 +1,135 @@
...
+prefix vann: <http://purl.org/vocab/vann/>
+prefix skos: <http://www.w3.org/2004/02/skos/core#>
+prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+prefix owl: <http://www.w3.org/2002/07/owl#>
+prefix xsd: <http://www.w3.org/2001/XMLSchema#>
+prefix binding: <https://www.w3.org/2019/wot/binding#>
+prefix hctl: <https://www.w3.org/2019/wot/hypermedia#>
+prefix sh: <http://www.w3.org/ns/shacl#>
+prefix coap: <https://www.w3.org/2019/wot/coap#>
```
ektrah, line 13: Should the prefix be `cov:` for consistency?

```suggestion
prefix cov: <https://www.w3.org/2019/wot/coap#>
```

ontology/coap.ttl
```diff
@@ -0,0 +1,165 @@
+@prefix : <https://www.w3.org/2019/wot/coap#> .
```
ektrah, line 1: Should the prefix be `cov:` for consistency?

```suggestion
@prefix cov: <https://www.w3.org/2019/wot/coap#> .
```

ektrah (changes requested):

ontology/coap.ttl
JKRhb: Thank you for your feedback on this. I took over the old ontology title for now, for lack of a better alternative. Would something like "CoAP Binding Ontology" be more appropriate? Or do you have something in mind that would be a better fit here?

ektrah: Since the vocabulary is already fully documented in `bindings/protocols/coap/index.html`, why not just move `ontology/coap.ttl` to `bindings/protocols/coap/coap.ttl` and delete `ontology/coap.html`?

bindings/protocols/coap/mapping.ttl
```diff
@@ -0,0 +1,87 @@
...
+        sh:maxCount 1 ;
+    ] ;
+    sh:property [
+        sh:path cov:hasAccept ;
+        sh:datatype xsd:unsignedShort ;
+        sh:maxCount 1 ;
+    ] .
+
+:CoapResponseShape a sh:NodeShape ;
+    sh:targetClass hctl:Form ;
```
ektrah, line 65: Is that the right target class?

bindings/protocols/coap/mapping.ttl
```diff
@@ -0,0 +1,87 @@
...
+	sh:order 2 ; # Used to select the shape
+    sh:closed false ;
+    sh:property [
+        sh:path cov:hasContentFormat ;
+        sh:datatype xsd:unsignedShort ;
+        sh:maxCount 1 ;
+    ] .
+
+:CoapBlockwiseShape a sh:NodeShape ;
+    sh:targetClass hctl:Form ;
```
ektrah, line 75: Is that the right target class?

bindings/protocols/coap/context.jsonld
```diff
@@ -0,0 +1,43 @@
+{
+    "@context": {
+        "xsd": "http://www.w3.org/2001/XMLSchema#",
+        "cov": "https://www.w3.org/2019/wot/coap#",
+        "method": {
+            "@id": "cov:hasMethod",
```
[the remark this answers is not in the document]
JKRhb: To be honest, I mostly tried to replicate what was done for Modbus and MQTT in this context file – I guess for the future, the process of creating the JSON-LD @context file for a protocol binding needs to be better documented and/or automated, especially for making it easier for third parties to create a WoT binding.

From my understanding, though, the @context simply links the vocabulary terms (e.g., `cov:method`) to ontology terms (in this case `https://www.w3.org/2019/wot/coap#hasMethod`). So I think this @context does not imply that a form itself is having a method or that it becomes a Request instance (otherwise something similar would also be true for MQTT and Modbus), but I might as well be wrong here. I will take a deeper dive into JSON-LD, but also RDF and SHACL to be able to address your concerns, since some points are still a bit unclear to me as well.

bindings/protocols/coap/mapping.ttl
JKRhb: Thank you, good catch :) Changed the target class to `ExpectedResponse` in e746e1f5048177b1c96164a828ab221a26adc35b. Since these definitions also apply to the `AdditionalExpectedResponse` class, I also added it as a target class. I suppose the document does not need a separate table for this definition, does it?

bindings/protocols/coap/mapping.ttl
JKRhb: Hmm, I am not entirely sure about this one. However, the definitions are located at the form level, so I assume it should be correct? Or do you see a better fit here?

ontology/coap.ttl
JKRhb: I think this change would require adding `cov:` to all definitions defined in this document, so for simplicity (and to be consistent with the other ontologies, where the same approach is used) I think I would keep it as is.

ontology/coap.ttl
```diff
@@ -0,0 +1,164 @@
...
+        rdfs:domain :Request ;
+        rdfs:comment "Indicates that a Consumer must include an Accept option [[RFC7252]] with the given value when requesting the resource associated with this Form."@en ;
+        rdfs:isDefinedBy <> ;
+        rdfs:label "hasAccept"@en .
+
+#################################################################
+#    Data properties
+#################################################################
+
+:BlockWiseTransferParameters rdf:type owl:DatatypeProperty ;
```
ektrah, line 79: Is `:BlockWiseTransferParameters` intended to be a datatype instead of a datatype property?

```suggestion
:BlockWiseTransferParameters rdf:type rdfs:Datatype ;
```

What's the value space of `:BlockWiseTransferParameters` literals? Wouldn't a class make more sense?

JKRhb: Good point, updated in https://github.com/w3c/wot-binding-templates/commit/4e29228500359ab12907dbe0e70c248a5a850742

bindings/protocols/coap/mapping.ttl
JKRhb: I replaced the target class now with `cov:BlockWiseTransferParameters` in 153f06e4c99796b368f79ad4db77bf2e38961e85. Does that make more sense?

ontology/coap.ttl
JKRhb: I was struggling a bit to reproduce the previous document state. Setting the datatype to `:Blocksize` in `mapping.ttl` replaces the `unsignedShort` with `Blocksize`, therefore I arrived at the current solution/workaround – but I suppose that the replacement would probably make more sense than mixing an enum data type with a "primitive" one, right?

egekorkan:
Call of 01.03:
- We should keep the modularity and the TTL files. The corresponding HTML is needed for linking but this should not be mandatory.
- The terms that are not used by our binding should not be mandatory (no full protocol ontology is needed).
- Ontologies should be ideally designed by the organization specifying the protocol. In some cases (like Modbus), we have to do it ourselves (and thus host it). We should look into other ontology work within W3C.

ontology/coap.ttl
ektrah: The HTML document for this vocabulary is `bindings/protocols/coap/index.html`. And [the title is `Web of Things (WoT) CoAP Binding Template`](https://github.com/w3c/wot-binding-templates/blob/5a681062759b023b2ed3a1129a2df5b6dde8c0f8/bindings/protocols/coap/index.html#L51). (Although I'd agree that "CoAP Vocabulary for the Web of Things" would probably be a better title for both the vocabulary and `bindings/protocols/coap/index.html`.)

bindings/protocols/coap/mapping.ttl
ektrah: Looks good. I agree the document does not need two separate tables for `ExpectedResponse` and `AdditionalExpectedResponse`.

bindings/protocols/coap/mapping.ttl
ektrah: `sh:targetClass cov:BlockWiseTransferParameters` looks good to me. However, `:BlockWiseTransferParameters` in coap.ttl still doesn't look right; see below.

ontology/coap.ttl
ektrah: `:BlockWiseTransferParameters` in coap.ttl now looks like this:

https://github.com/w3c/wot-binding-templates/blob/e71a3954aecd94bd4f45ca0eae3c263b7c1282fa/ontology/coap.ttl#L122-L126

It's defined both as an instance of `owl:Class` (line 122) and to have a domain (line 123), which makes it an instance of `rdfs:Property`. Should line 123 be removed´?

Also, should `:hasBlockwise` and `:hasQBlockwise` have `:BlockWiseTransferParameters` as range?

ontology/coap.ttl
ektrah: I think I'd add it, but that's up to you.

ontology/coap.ttl
ektrah: I think I'd remove the `:Blocksize` class and instances from coap.ttl and just use `xsd:unsignedShort`. After all, I don't see mapping.ttl doing any actual mapping; it's just validating that the triples in a TD have certain shape, similar to how the JSON Schema is validating that the JSON objects in the TD follow a certain schema.

ektrah:
Thanks a lot @JKRhb for addressing all my comments! It's a big step in the right direction, I think.

Overall, however, I think there needs to be a bit more work done as it's unclear how what the @context file, `coap.ttl`, and `mapping.ttl` should exactly look like to make everything work. This may be something to discuss in one of the upcoming calls.

For instance, if you copy this example into the [JSON-LD playground](https://json-ld.org/playground/)...

```json
{
  "@context": [
    "https://www.w3.org/2022/wot/td/v1.1",
    {
      "xsd": "http://www.w3.org/2001/XMLSchema#",
      "cov": "https://www.w3.org/2019/wot/coap#",
      "method": {
        "@id": "cov:hasMethod",
        "@type": "xsd:string"
      },
      "BlockWiseTransferParameters": {
        "@id": "cov:BlockWiseTransferParameters"
      },
      "blockwise": {
        "@id": "cov:hasBlockwise",
        "@type": "@id",
        "@container": "@index"
      },
      "qblockwise": {
        "@id": "cov:hasQBlockwise",
        "@type": "@id",
        "@container": "@index"
      },
      "block2SZX": {
        "@id": "cov:hasblock2SZX",
        "@type": "xsd:unsignedShort"
      },
      "block1SZX": {
        "@id": "cov:hasblock1SZX",
        "@type": "xsd:unsignedShort"
      },
      "accept": {
        "@id": "cov:hasAccept",
        "@type": "xsd:unsignedShort"
      },
      "contentFormat": {
        "@id": "cov:hasContentFormat",
        "@type": "xsd:unsignedShort"
      },
      "hopLimit": {
        "@id": "cov:hasHopLimit",
        "@type": "xsd:unsignedByte"
      }
    }
  ],
  "actions": {
    "start": {
      "input": { "type": "integer" },
      "output": { "type": "string" },
      "forms": [
        {
          "href": "coap://[2001:DB8::1]/start",
          "cov:method": "POST",
          "cov:accept": 60,
          "response": {
            "contentType": "application/cbor",
            "cov:contentFormat": 60
          }
        },
        {
          "href": "coap://[2001:DB8::1]/start",
          "cov:method": "POST",
          "contentType": "application/cbor",
          "cov:contentFormat": 60,
          "cov:accept": 50,
          "response": {
            "contentType": "application/json",
            "cov:contentFormat": 50
          }
        }
      ]
    }
  }
}
```

...it gets expanded to this...

```json
[
  {
    "https://www.w3.org/2019/wot/td#hasActionAffordance": [
      {
        "https://www.w3.org/2019/wot/td#hasForm": [
          {
            "https://www.w3.org/2019/wot/coap#accept": [
              {
                "@value": 60
              }
            ],
            "https://www.w3.org/2019/wot/coap#method": [
              {
                "@value": "POST"
              }
            ],
            "https://www.w3.org/2019/wot/hypermedia#hasTarget": [
              {
                "@type": "http://www.w3.org/2001/XMLSchema#anyURI",
                "@value": "coap://[2001:DB8::1]/start"
              }
            ],
            "https://www.w3.org/2019/wot/hypermedia#returns": [
              {
                "https://www.w3.org/2019/wot/hypermedia#forContentType": [
                  {
                    "@value": "application/cbor"
                  }
                ],
                "https://www.w3.org/2019/wot/coap#contentFormat": [
                  {
                    "@value": 60
                  }
                ]
              }
            ]
          },
          {
            "https://www.w3.org/2019/wot/hypermedia#forContentType": [
              {
                "@value": "application/cbor"
              }
            ],
            "https://www.w3.org/2019/wot/coap#accept": [
              {
                "@value": 50
              }
            ],
            "https://www.w3.org/2019/wot/coap#contentFormat": [
              {
                "@value": 60
              }
            ],
            "https://www.w3.org/2019/wot/coap#method": [
              {
                "@value": "POST"
              }
            ],
            "https://www.w3.org/2019/wot/hypermedia#hasTarget": [
              {
                "@type": "http://www.w3.org/2001/XMLSchema#anyURI",
                "@value": "coap://[2001:DB8::1]/start"
              }
            ],
            "https://www.w3.org/2019/wot/hypermedia#returns": [
              {
                "https://www.w3.org/2019/wot/hypermedia#forContentType": [
                  {
                    "@value": "application/json"
                  }
                ],
                "https://www.w3.org/2019/wot/coap#contentFormat": [
                  {
                    "@value": 50
                  }
                ]
              }
            ]
          }
        ],
        "https://www.w3.org/2019/wot/td#hasInputSchema": [
          {
            "@type": [
              "https://www.w3.org/2019/wot/json-schema#IntegerSchema"
            ]
          }
        ],
        "https://www.w3.org/2019/wot/td#hasOutputSchema": [
          {
            "@type": [
              "https://www.w3.org/2019/wot/json-schema#StringSchema"
            ]
          }
        ],
        "https://www.w3.org/2019/wot/td#name": [
          {
            "@value": "start"
          }
        ]
      }
    ]
  }
]
```

...which doesn't look right to me.

Also, I think that the vocabulary documented in `bindings/protocols/coap/index.html` is misplaced in `ontology/coap.ttl` (should be `bindings/protocols/coap/coap.ttl`) and that `ontology/coap.html` is superfluous since the vocabulary is already documented in `bindings/protocols/coap/index.html`.

egekorkan:
For previews, we can use https://deploy-preview-246--wot-binding-templates.netlify.app/bindings/protocols/coap/index.html

ektrah:
@JKRhb Sorry for the holdup ☹️

relu91:
> The discussion on whether we should have HTML files of the ontologies will continue on 22.03. @relu91 would be nice if you are available.

I will!✅

bindings/protocols/coap/mapping.ttl
```diff
@@ -0,0 +1,108 @@
...
+@prefix schema: <http://schema.org/> .
+@prefix sh: <http://www.w3.org/ns/shacl#> .
+@prefix cov: <https://www.w3.org/2019/wot/coap#> .
+@prefix binding: <https://www.w3.org/2019/wot/binding#> .
+@base <https://www.w3.org/ns/wot/binding/coap-validation#> .
+
+<> rdf:type owl:Ontology ;
+    owl:imports <https://www.w3.org/2019/wot/coap> ;
+    vann:preferredNamespacePrefix "cov" ;
+    vann:preferredNamespaceUri : ;
```
ektrah, line 25:
```suggestion
    vann:preferredNamespaceUri "https://www.w3.org/ns/wot/binding/coap-validation#" ;
```

bindings/protocols/coap/mapping.ttl
```diff
@@ -0,0 +1,108 @@
...
+    rdfs:comment "Vocabulary to map CoAP concepts to the standard Thing Interaction model."@en ;
+    rdfs:label "CoAP protocol binding"@en .
+
+:CoapFormShape a sh:NodeShape ;
+    sh:targetClass hctl:Form ;
+	sh:order 1 ; # Used to select the shape
```
ektrah, line 31:
```suggestion
    sh:order 1 ; # Used to select the shape
```

bindings/protocols/coap/mapping.ttl
```diff
@@ -0,0 +1,108 @@
...
+@prefix sh: <http://www.w3.org/ns/shacl#> .
+@prefix cov: <https://www.w3.org/2019/wot/coap#> .
+@prefix binding: <https://www.w3.org/2019/wot/binding#> .
+@base <https://www.w3.org/ns/wot/binding/coap-validation#> .
+
+<> rdf:type owl:Ontology ;
+    owl:imports <https://www.w3.org/2019/wot/coap> ;
+    vann:preferredNamespacePrefix "cov" ;
+    vann:preferredNamespaceUri : ;
+    rdfs:comment "Vocabulary to map CoAP concepts to the standard Thing Interaction model."@en ;
```
ektrah, line 26: Does it, though?

bindings/protocols/coap/mapping.ttl
```diff
@@ -0,0 +1,108 @@
...
+    rdfs:label "CoAP protocol binding"@en .
+
+:CoapFormShape a sh:NodeShape ;
+    sh:targetClass hctl:Form ;
+	sh:order 1 ; # Used to select the shape
+    sh:closed false ;
+    sh:property [
+        sh:path cov:hasMethod ;
```
ektrah, line 34:
```suggestion
        sh:path cov:method ;
```

bindings/protocols/coap/mapping.ttl
```diff
@@ -0,0 +1,108 @@
...
+        sh:datatype xsd:string ;
+        sh:maxCount 1 ;
+    ] ;
+    sh:property [
+        sh:path cov:hasBlockwise ;
```
ektrah, line 39:
```suggestion
        sh:path cov:blockwise ;
```

bindings/protocols/coap/mapping.ttl
```diff
@@ -0,0 +1,108 @@
...
+        sh:datatype cov:BlockWiseTransferParameters ;
```
ektrah, line 40:
```suggestion
        sh:class cov:BlockWiseTransferParameters ;
```

bindings/protocols/coap/mapping.ttl
```diff
@@ -0,0 +1,108 @@
...
+        sh:maxCount 1 ;
+    ] ;
+    sh:property [
+        sh:path cov:hasQBlockwise ;
```
ektrah, line 44:
```suggestion
        sh:path cov:qblockwise ;
```

bindings/protocols/coap/mapping.ttl
```diff
@@ -0,0 +1,108 @@
...
+        sh:datatype cov:BlockWiseTransferParameters ;
```
ektrah, line 45:
```suggestion
        sh:class cov:BlockWiseTransferParameters ;
```

bindings/protocols/coap/mapping.ttl
```diff
@@ -0,0 +1,108 @@
...
+        sh:maxCount 1 ;
+    ] ;
+    sh:property [
+        sh:path cov:hasHopLimit ;
```
ektrah, line 49:
```suggestion
        sh:path cov:hopLimit ;
```

bindings/protocols/coap/mapping.ttl
```diff
@@ -0,0 +1,108 @@
...
+        sh:datatype xsd:unsignedByte ;
+        sh:maxCount 1 ;
+    ] ;
+    sh:property [
+        sh:path cov:hasContentFormat ;
```
ektrah, line 54:
```suggestion
        sh:path cov:contentFormat ;
```

bindings/protocols/coap/mapping.ttl
```diff
@@ -0,0 +1,108 @@
...
+        sh:datatype xsd:unsignedShort ;
+        sh:maxCount 1 ;
+    ] ;
+    sh:property [
+        sh:path cov:hasAccept ;
```
ektrah, line 59:
```suggestion
        sh:path cov:accept ;
```

bindings/protocols/coap/mapping.ttl
```diff
@@ -0,0 +1,108 @@
...
+        sh:datatype xsd:unsignedShort ;
+        sh:maxCount 1 ;
+    ] .
+
+:CoapResponseShape a sh:NodeShape ;
+    sh:targetClass hctl:ExpectedResponse, hctl:AdditionalExpectedResponse ;
+    sh:order 2 ; # Used to select the shape
+    sh:closed false ;
+    sh:property [
+        sh:path cov:hasContentFormat ;
```
ektrah, line 69:
```suggestion
        sh:path cov:contentFormat ;
```

bindings/protocols/coap/mapping.ttl
```diff
@@ -0,0 +1,108 @@
...
+        sh:maxCount 1 ;
+    ] .
+
+:CoapBlockwiseShape a sh:NodeShape ;
+    sh:targetClass cov:BlockWiseTransferParameters;
+    sh:order 3 ; # Used to select the shape
+    sh:closed false ;
+    sh:property [
+        sh:nodeKind sh:Literal ;
+        sh:path cov:hasBlock2SZX ;
```
ektrah, line 80:
```suggestion
        sh:path cov:block2SZX ;
```

bindings/protocols/coap/mapping.ttl
```diff
@@ -0,0 +1,108 @@
...
+            64
+            128
+            256
+            512
+            1024
+        ) ;
+    ] ;
+    sh:property [
+        sh:nodeKind sh:Literal ;
+        sh:path cov:hasBlock1SZX ;
```
ektrah, line 95:
```suggestion
        sh:path cov:block1SZX ;
```

bindings/protocols/coap/coap.ttl
```diff
@@ -0,0 +1,99 @@
...
+        rdfs:label "hasBlock1SZX"@en .
+
+#################################################################
+#    Classes
+#################################################################
+
+:BlockWiseTransferParameters rdf:type rdfs:Class, owl:Class ;
+        rdfs:isDefinedBy <> ;
+        rdfs:comment "Parameters used for block-wise transfer [[RFC7959]] and block-wise transfer supporting robust transmission [[RFC9177]]."@en ;
+        rdfs:label "BlockWiseTransferParameters"@en .
```
ektrah, line 99:
```suggestion
@prefix cov: <https://www.w3.org/2019/wot/coap#> .
@prefix dcterms: <http://purl.org/dc/terms/> .
@prefix owl: <http://www.w3.org/2002/07/owl#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix schema: <http://schema.org/> .
@prefix vann: <http://purl.org/vocab/vann/> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
@base <https://www.w3.org/2019/wot/coap> .

<> a owl:Ontology ;
   dcterms:title "CoAP Vocabulary for the Web of Things"@en ;
   dcterms:creator [ a dcterms:Agent ,
                       schema:Person ;
                     schema:name "Klaus Hartke" ;
                     schema:affiliation [ a schema:Organization ;
                                          schema:name "Siemens AG" ;
                                          schema:url "https://www.siemens.com/"
                                        ]
                   ] ,
                   [ a dcterms:Agent ,
                       schema:Person ;
                     schema:name "Jan Romann" ;
                     schema:affiliation [ a schema:Organization ;
                                          schema:name "Invited Expert" ;
                                          schema:url "https://github.com/JKRhb/"
                                        ]
                   ] ;
   vann:preferredNamespacePrefix "cov" ;
   vann:preferredNamespaceUri "https://www.w3.org/2019/wot/coap#" .

cov:method a rdf:Property ;
           rdfs:comment "Indicates the CoAP method to be used in the request."@en ;
           rdfs:isDefinedBy <> ;
           rdfs:seeAlso "See the <a href=\"https://www.iana.org/assignments/core-parameters/#method-codes\">CoAP Method Codes</a> registry for the set of possible values.";
           rdfs:label "method"@en .

cov:blockwise a rdf:Property ;
              rdfs:range cov:BlockWiseTransferParameters ;
              rdfs:comment "Indicates that block-wise transfer [[RFC7959]] is supported."@en ;
              rdfs:isDefinedBy <> ;
              rdfs:label "blockwise"@en .

cov:qblockwise a rdf:Property ;
               rdfs:range cov:BlockWiseTransferParameters ;
               rdfs:comment "Indicates that quick block-wise transfer [[RFC9177]] is supported."@en ;
               rdfs:isDefinedBy <> ;
               rdfs:label "qblockwise"@en .

cov:hopLimit a rdf:Property ;
             rdfs:comment "Indicates that a CoAP Hop-Limit option [[RFC8768]] should be included in the request."@en ;
             rdfs:isDefinedBy <> ;
             rdfs:label "hopLimit"@en .

cov:contentFormat a rdf:Property ;
                  rdfs:comment "Indicates a numeric CoAP Content-Format [[RFC7252]] as a combination of Media Type and Content Coding. Must match the given <code>contentType</code> and <code>contentCoding</code> members."@en ;
                  rdfs:isDefinedBy <> ;
                  rdfs:label "contentFormat"@en .

cov:accept a rdf:Property ;
           rdfs:comment "Indicates that a Consumer must include an Accept option [[RFC7252]] with the given value when requesting the resource associated with this Form."@en ;
           rdfs:isDefinedBy <> ;
           rdfs:label "accept"@en .

cov:block2SZX a rdf:Property ;
              rdfs:domain cov:BlockWiseTransferParameters ;
              rdfs:range xsd:integer ;
              rdfs:comment "Indicates the largest block size that may be used in a Block2 or Q-Block2 Option"@en ;
              rdfs:label "block2SZX"@en .

cov:block1SZX a rdf:Property ;
              rdfs:domain cov:BlockWiseTransferParameters ;
              rdfs:range xsd:integer ;
              rdfs:comment "Indicates the largest block size that may be used in a Block1 or Q-Block1 Option"@en ;
              rdfs:label "block1SZX"@en .

cov:BlockWiseTransferParameters a rdfs:Class ;
                                rdfs:isDefinedBy <> ;
                                rdfs:comment "Parameters used for block-wise transfer [[RFC7959]] and block-wise transfer supporting robust transmission [[RFC9177]]."@en ;
                                rdfs:label "BlockWiseTransferParameters"@en .
```

ektrah (changes requested):

bindings/protocols/coap/mapping.ttl
ektrah: I might be good to have some "unit tests" in the form of some TDs that get validated against these SHACL shapes as part of CI/CD.

bindings/protocols/coap/coap.ttl
```diff
@@ -0,0 +1,90 @@
...
+@prefix schema: <http://schema.org/> .
+@prefix vann: <http://purl.org/vocab/vann/> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+@base <https://www.w3.org/2019/wot/coap> .
+
+<> a owl:Ontology ;
+   dcterms:title "CoAP Vocabulary for the Web of Things"@en ;
+   rdfs:label "CoAP Vocabulary for the Web of Things"@en ; # TODO: Refactor
+   # TODO: Also add dctrms:creator?
+   dcterms:author [ a dcterms:Agent ,
```
ektrah, line 15: There is no `dcterms:author` in [DCMI Metadata Terms](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/).

bindings/protocols/coap/context.jsonld
```diff
@@ -0,0 +1,38 @@
...
+        "method": {
+            "@id": "cov:method",
+            "@type": "xsd:string"
+        },
+        "blockwise": {
+            "@id": "cov:blockwise",
+            "@type": "@vocab"
+        },
+        "qblockwise": {
+            "@id": "cov:qBlockwise",
```
ektrah, line 14:
```suggestion
            "@id": "cov:qblockwise",
```

bindings/protocols/coap/coap.ttl
```diff
@@ -0,0 +1,89 @@
...
+                   [ a dcterms:Agent ,
+                       schema:Person ;
+                     schema:name "Jan Romann" ;
+                     schema:affiliation [ a schema:Organization ;
+                                          schema:name "Invited Expert" ;
+                                          schema:url "https://github.com/JKRhb/"
+                                        ]
+                   ] ;
+   vann:preferredNamespacePrefix "cov" ;
+   vann:preferredNamespaceUri cov: ;
```
ektrah, line 31:
```suggestion
   vann:preferredNamespaceUri "https://www.w3.org/2019/wot/coap#" ;
```

The value of `vann:preferredNamespaceUri` should probably be a string. [VANN doesn't say so explicitly](https://vocab.org/vann/) but it's [how VANN uses it itself](https://vocab.org/vann/vann-vocab-20100607.rdf). It also makes more sense than an IRI since `https://www.w3.org/2019/wot/coap#` does not refer to a node in the RDF graph here but is intended to be prepended to names with the `cov:` prefix.

bindings/protocols/coap/context.jsonld
```diff
@@ -0,0 +1,38 @@
...
+            "@type": "xsd:string"
+        },
+        "blockwise": {
+            "@id": "cov:blockwise",
+            "@type": "@vocab"
+        },
+        "qblockwise": {
+            "@id": "cov:qblockwise",
+            "@type": "@vocab"
+        },
```
ektrah, line 16: Should this be referring to the `BlockWiseTransferParameters` class somehow?

ektrah: I've tried a few variants in the JSON-LD playground but couldn't get it to emit a `"@type": "cov:BlockWiseTransferParameters"` if there is no `@type` in the data. Setting the `@type` to a class in an @context doesn't seem to be doing anything.

<details>
<summary>Details (click to expand)</summary>

```turtle
{
  "@context": {
    "cov": "https://www.w3.org/2019/wot/coap#",
    "hctl": "https://www.w3.org/2019/wot/hypermedia#",
    "xsd": "http://www.w3.org/2001/XMLSchema#",
    "Form": {
      "@id": "hctl:Form",
      "@context": {
        "blockwise": {
          "@id": "cov:blockwise"
        }
      }
    },
    "BlockWiseTransferParameters": {
      "@id": "cov:BlockWiseTransferParameters",
      "@context": {
        "blockSize1": {
          "@id": "cov:blockSize1",
          "@type": "xsd:unsignedShort"
        }
      }
    }
  },
  "@type": "Form",
  "blockwise": {
    "@type": "BlockWiseTransferParameters",
    "blockSize1": 24
  }
}
```

```turtle
{
  "@context": {
    "cov": "https://www.w3.org/2019/wot/coap#",
    "hctl": "https://www.w3.org/2019/wot/hypermedia#",
    "xsd": "http://www.w3.org/2001/XMLSchema#",
    "Form": {
      "@id": "hctl:Form",
      "@context": {
        "blockwise": {
          "@id": "cov:blockwise",
          "@context": {
            "blockSize1": {
              "@id": "cov:blockSize1",
              "@type": "xsd:unsignedShort"
            }
          }
        }
      }
    }
  },
  "@type": "Form",
  "blockwise": {
    "blockSize1": 24
  }
}
```

</details>

Having `"@type": "@vocab"`` or not also doesn't seem to make a difference. For reference, I found [this in the JSON-LD spec](https://www.w3.org/TR/json-ld/#included-nodes), which I think is not what we're doing here.

So, I guess the best we can do is this (and people will need to apply `coap.ttl` after converting a TD to triples in order to infer the type of the `cov:blockwise` property value):

```suggestion
        "blockwise": {
            "@id": "cov:blockwise"
        },
        "qblockwise": {
            "@id": "cov:qblockwise"
        },
```

ektrah (approved):
Open issues:

* How to use the context file in a TD?
    * Does <https://www.w3.org/2022/wot/td/v1.1> need to be updated?
    * Shouldn't it be `cov:method` rather than `method` in a TD?
    * Would a simple `"cov": "https://www.w3.org/2019/wot/coap#"` suffice?
* What's the role of the mapping file as it's actually not mapping anything?
    * Should it be renamed to `validation.ttl`?
    * Could it be merged into `coap.ttl`?
* Do we really need `coap.html`?
    * The vocabulary is already documented in `index.html`
    * Why have another file that comes without any explanations or examples?

For me, the PR looks now good enough that it could be merged and these open issues be addressed in a subsequent PR.

bindings/protocols/coap/mapping.ttl
ektrah: Follow-up PR is OK

relu91 (approved):
PR follows the requirements that we defined for a binding template subspecification. I remember that we discussed the reasons why the ontology file is now under the protocol, but I feel that we need to either explain this decision somewhere (readme?) or still move it under the ontology folder as the others (or move the others to their protocol folder)

JKRhb:
> PR follows the requirements that we defined for a binding template subspecification. I remember that we discussed the reasons why the ontology file is now under the protocol, but I feel that we need to either explain this decision somewhere (readme?) or still move it under the ontology folder as the others (or move the others to their protocol folder)

Hmm, good point, I now moved the ontology files back to the ontology folder. Moving all the ontologies to their respective protocol binding folders could then be done in a follow-up PR in order to better document the decision.

egekorkan:
Call of 22.03:
- The group agrees to merge with 3 remaining comments to be addressed later on:
 - https://github.com/w3c/wot-binding-templates/pull/246/#discussion_r1133131871
 - https://github.com/w3c/wot-binding-templates/pull/246/#discussion_r1133125585
 - https://github.com/w3c/wot-binding-templates/pull/246/#pullrequestreview-1333140744
- We will do folder restructuring for moving ontologies into the bindings folders

PR #260, opened 2023-03-08
Use `dct:creator` instead of `dct:author` in ontologies

JKRhb:
@ektrah noticed in https://github.com/w3c/wot-binding-templates/pull/246#discussion_r1129979924 that the term `author` that is currently being used to identify the authors of an ontology does not actually exist in the [DCMI Metadata Terms](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/). However, the term `creator` can be used instead.

This PR resolves the issue by simply replacing `dct:author` with `dct:creator`, leaving the resulting rendered ontologies unchanged.

### DIFF
--- a/ontology/modbus.template.rq
+++ b/ontology/modbus.template.rq
@@ -69,7 +69,7 @@ template :editor(?o) {
     "companyURL:" ?url
     "}" ; separator = ","
 } where {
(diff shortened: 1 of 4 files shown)


relu91 (approved):
Nice catch.

egekorkan (approved):

egekorkan:
Call of 15.03: Merging

Issue #238, opened 2023-02-08
Clarifying Binding Mechanism


egekorkan:
From TD and Architecture, we should bring (copyover) the parts that are about the binding mechanism

egekorkan:
Actually, the scripting API explains how a binding should be used in software. We should use the information there.

https://w3c.github.io/wot-scripting-api/#the-readproperty-method

egekorkan:
From the call of 15.02, some important notes: 
- We should state that a protocol binding cannot be used on its own since serialization is a MUST by the TD spec (requiring `contentType`). 
- Explain what happens when an ecosystem binding is needed and why it is needed.

relu91:
Linking Scripting issue here: https://github.com/w3c/wot-scripting-api/issues/430

egekorkan:
Regarding relationship to scripting: I think that a generic mechanism needs to be explained here. Not all implementations have the separation of a runtime (servient and its bindings) and the developer code. So scripting will need a more clear definition of where this separation happens than the bindings template. I think Scripting is already doing this well in the algorithms for methods like `readproperty()`

egekorkan:
Relevant parts from TD:
- Overall meaning: https://w3c.github.io/wot-thing-description/#form the paragraph at the intro
- https://w3c.github.io/wot-thing-description/#sec-op-data-schema-mapping for overall understanding of data schema and operation relationship
- HTTP Binding: https://w3c.github.io/wot-thing-description/#http-binding-assertions
- Examples for CoAP and MQTT: https://w3c.github.io/wot-thing-description/#myLampThing-example-serialization

Relevant parts from Architecture:
- Hypermedia principle: https://w3c.github.io/wot-architecture/#sec-affordances
- Protocol Bindings section: https://w3c.github.io/wot-architecture/#sec-protocol-bindings
- Media Types section: https://w3c.github.io/wot-architecture/#media-types
- Binding Templates building block: https://w3c.github.io/wot-architecture/#sec-binding-templates -> This has a nice diagram that should be considered.
- TDs for existing devices: https://w3c.github.io/wot-architecture/#existing-impl

If I have forgot about something, I will update the lists above based on the comments.

PR #262, opened 2023-03-15
Explain core binding mechanism

### LINKED ISSUE #252: Mention that a binding is mandatory for a TD to exist
Coming from https://github.com/w3c/wot-binding-templates/issues/238#issuecomment-1431640818

### LINKED ISSUE #255: A diagram to help understand what is happening in operation mappings.

egekorkan:
fixes #255 and starts addressing #238
fixes #252

[Netlify Preview](https://deploy-preview-262--wot-binding-templates.netlify.app/)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-binding-templates/pull/262.html" title="Last updated on Mar 22, 2023, 9:09 AM UTC (3210590)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-binding-templates/262/6e03f10...3210590.html" title="Last updated on Mar 22, 2023, 9:09 AM UTC (3210590)">Diff</a>


egekorkan:
Call of 15.03:
- Overall the diagram is good
- An SVG version would be better to avoid the resolution problem
- It would be good to have a sequence diagram with protocol requests for each protocol binding
- A sentence before the diagram to explain the application logic would be nice (here it would be, "this is how a message sequence would look like if the Consumer wants the robot to move to x:12 and y:100". However, it would be better take one of the existing use cases (e.g. air conditioner)

danielpeintner:
> * An SVG version would be better to avoid the resolution problem.

I agree with using SVG.
I just wonder why in the current TD document we use PNG instead of the SVG counterpart (they exist in most of the cases). Shall we open another issue/PR to change that also?

egekorkan:
> I just wonder why in the current TD document we use PNG instead of the SVG counterpart

I am not 100% sure but SVG is not the best when it comes to text representation. The automatically generated SVGs are converted to PNGs at some point.

egekorkan:
I have taken all the feedback from the last call and everything looks nicer now and has more explanations.

danielpeintner:
I noticed you are using quotes for JSON terms now 👍
I also noticed that for `"input": { x: ..., y: ...}` you are not doing that. Just an oversight?

egekorkan:
@danielpeintner it was definitely an oversight, I have changed it now.

danielpeintner:
> I have changed it now.

I think you missed to update the PNG version. Anyhow, didn't we want to remove PNG anyway?

egekorkan:
> > I have changed it now.
> 
> I think you missed to update the PNG version. Anyhow, didn't we want to remove PNG anyway?

Good catch. I have removed the old and did the same with the other figure we have.

relu91 (approved):

egekorkan:
Call of 22.03: No objections, merging

PR #269, opened 2023-03-22
Move ontologies into individual protocol binding folders

JKRhb:
As discussed in the context of #246, this PR moves all ontology-related files to the respective protocol binding folders and adjusts the documents linking to them.

***
<a href="https://pr-preview.s3.amazonaws.com/JKRhb/wot-binding-templates/pull/269.html" title="Last updated on Mar 22, 2023, 9:29 PM UTC (fe5c99b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-binding-templates/269/c314cbd...JKRhb:fe5c99b.html" title="Last updated on Mar 22, 2023, 9:29 PM UTC (fe5c99b)">Diff</a>


bindings/protocols/modbus/modbus.template.rq
JKRhb, line 1: While moving it, I noticed that this file is not actually being used. Should we still keep it? Or could it also be removed?

egekorkan: @relu91 could you have a look?

egekorkan (approved):

egekorkan:
Also realized that it will make the publication process easier and will give each collaborator its own folder to manage and be responsible of.

Realized after approval, could you adapt the https://github.com/w3c/wot-binding-templates/blob/main/.github/CODEOWNERS file?

ektrah (approved):

egekorkan:
Call of 29.03: 
- This is the right direction to have "packages" per protocol
- In the readme and vocab creation guide, we should explain the difference between different types of ontologies (our CoAP vs HTTP in RDF)
- We should clarify how we publish the ontology to the public under the w3.org (content negotiation and other possible technical implications)
- Merging

PR #276, opened 2023-03-26
fix(coap): clarify purpose of cov:contentFormat

### LINKED ISSUE #257: Clarify cov:contentFormat
https://github.com/w3c/wot-binding-templates/blob/5a681062759b023b2ed3a1129a2df5b6dde8c0f8/bindings/protocols/coap/index.html#L509

This doesn't tell why anyone would want to indicate a CoAP Content-Format in a form. We should clarify this.

JKRhb:
This PR attempts to clarify the purpose of the term `cov:contentFormat` in the CoAP Protocol Binding document and to fix #257. This is only a first try, any suggestions for improvements to the wording are very welcome :)


egekorkan (approved):

ektrah:
IMHO this still doesn't provide any hint on the purpose of `cov:contentFormat` in a form, only says what a CoAP Content-Format is. (The description of `contentType` in the TD spec is equally unhelpful.)

JKRhb:
> IMHO this still doesn't provide any hint on the purpose of `cov:contentFormat` in a form, only says what a CoAP Content-Format is. (The description of `contentType` in the TD spec is equally unhelpful.)

Hmm, I was a bit unsure how detailed the description should be since we already have an explanation in the ["Content Negotiation" section](https://w3c.github.io/wot-binding-templates/bindings/protocols/coap/index.html#content-negotiation). I will try to reword the description a bit.

ontology/coap.ttl
```diff
@@ -58,7 +58,7 @@ cov:hopLimit a rdf:Property, owl:DatatypeProperty ;
 
 cov:contentFormat a rdf:Property, owl:DatatypeProperty ;
                   rdfs:range xsd:unsignedShort ;
-                  rdfs:comment "Indicates a numeric CoAP Content-Format [[RFC7252]] as a combination of Media Type and Content Coding. Must match the given <code>contentType</code> and <code>contentCoding</code> members."@en ;
+                  rdfs:comment "Indicates a numeric CoAP Content-Format [[RFC7252]] as a combination of Media Type and Content Coding as a guideline for Consumers without knowledge of the values registered with IANA. Must match the given <code>contentType</code> and <code>contentCoding</code> members."@en ;
```
JKRhb, line 61: Would you consider this an improvement? We could probably also add a reference to [appendix C](https://w3c.github.io/wot-thing-description/#contentType-usage) of the TD specification somewhere.

```suggestion
                  rdfs:comment "Indicates a numeric CoAP Content-Format [[RFC7252]] as a combination of Media Type and Content Coding as a guideline for Consumers without knowledge of the values registered with IANA. Based on this information, Consumers can choose an appropriate Form for an Interaction Affordance and then use the correct Content-Format in CoAP requests. Must match the given <code>contentType</code> and <code>contentCoding</code> members."@en ;
```

ektrah: That's a step in the right direction, I think. However, I'm not sure if a client would actually use `cov:contentFormat` to select the most appropriate form, or if the selection was made via `contentType` and the client then just used `cov:contentFormat` to construct the request. Perhaps a good next step would be to first clarify the purpose of `contentType` in a form. Is it really just for choosing the form that gives the client the best response, or does it also affect how the client should construct its request?

JKRhb: > Is it really just for choosing the form that gives the client the best response, or does it also affect how the client should construct its request?

It would also prescribe which content type/coding/format would be used for the request body/payload, so if a `contentType` of `application/cbor` was included in a form for an action that defines an `input` data schema, then the client would need to use CBOR for encoding the data that is required for invoking that action. I tried to express that in the second half of the suggestion above, but there is probably room for further clarification.

ektrah:
```suggestion
                  rdfs:comment "Prescribes which content type/coding/format is to be used for the request body/payload"@en ;
```

JKRhb: Hmm, I think that this would actually be an oversimplification since the purpose of contentType can vary based on the presence of, for instance, `input`, `output`, and `response` (also see the TD appendix cited above). I think you could argue that Things are a bit complicated here.

But maybe something like this could work? (I kept the relationship with `contentType` and `contentCoding` here.)

```suggestion
                  rdfs:comment "Prescribes the Content-Format to be used with CoAP request or response payloads. Must match the given <code>contentType</code> and <code>contentCoding</code> members in the same form."@en ;
```

ektrah: Yes, I think that's getting close! Thanks for putting so much thought into it. And I agree that it's difficult to capture the purpose of `contentType` in a concise sentence. But if we can't even explain it in a sentence, how can we expect readers to understand it? 😊 

Maybe like this?

```suggestion
                  rdfs:comment "Specifies the content format to be used by a client for input and the content format to be used by a server for output (where applicable). The value must be the numeric Content-Format ID registered with IANA for the combination of the <code>contentType</code> and <code>contentCoding</code> specified in the same form."@en ;
```

JKRhb: That looks very good! :) Thank you!

egekorkan: I really like this last solution, thanks for the cooperation @ektrah and @JKRhb !!

ektrah (approved):

egekorkan:
Call of 05.04: Merging.

[Preview](https://deploy-preview-276--wot-binding-templates.netlify.app/)

PR #296, opened 2023-09-21
Modbus: align the schema with the current ontology

### LINKED ISSUE #295: Modbus spec and schema does not match on terms
https://w3c.github.io/wot-binding-templates/bindings/protocols/modbus/index.html#form-terms has function but the schema has functionCode at https://github.com/w3c/wot-binding-templates/blob/main/bindings/protocols/modbus/modbus.schema.json#L22

relu91:
I updated the Modbus json schema accordingly to the lastest changes in the ontology and the index.html file. 

fix #295

### DIFF
--- a/bindings/protocols/modbus/modbus.schema.json
+++ b/bindings/protocols/modbus/modbus.schema.json
@@ -19,22 +19,8 @@
                         "InputRegister"
                     ]
                 },
-                "modbus:functionCode": {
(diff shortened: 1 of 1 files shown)


egekorkan:
Call of 27.09: changes are only to schema file, merging. If there are other issues, we will make other PRs.

PR #297, opened 2023-09-25
Modbus: move required array to modbusForm

alicanipek:
I updated the Modbus json schema and moved required array to modbusForm object. It was inside properties object but it should be inside modbusForm object.

### DIFF
--- a/bindings/protocols/modbus/modbus.schema.json
+++ b/bindings/protocols/modbus/modbus.schema.json
@@ -50,9 +50,9 @@
                 "modbus:quantity" : { "type": "integer", "minimum": 1},
                 "modbus:zeroBasedAddressing" : { "type" : "boolean"},
                 "modbus:timeout" : { "type": "number", "minimum": 0},
-                "modbus:unitID": { "type": "integer", "minimum": 0},
-                "required": ["modbus:address", "modbus:unitID"]
-            }
+                "modbus:unitID": { "type": "integer", "minimum": 0}
+            },
+            "required": ["modbus:address", "modbus:unitID"]
         },
         "affordance": {
             "type": "object",


relu91:
Thank you for your contribution, but you are working on an outdated version of the JSON schema. The correct one is coming from #296. Can you check if it works for you?

alicanipek:
Hi, I saw that pull request, but it doesn't have the required array. Can we say that those properties are no longer required?

relu91:
There has been quite of refactoring in the Modbus binding document. As you can [see](https://w3c.github.io/wot-binding-templates/bindings/protocols/modbus/index.html#form-terms), no form element is required anymore. The elements you are talking about have been moved to the [URL](https://w3c.github.io/wot-binding-templates/bindings/protocols/modbus/index.html#url).

hadjian:
Hi @relu91, thx for the hint. Is there a discussion/rational why this was moved to the URL? I liked the human readability of explicit attributes.

relu91:
I remember a looong discussion about these topics. I've found an [issue](https://github.com/w3c/wot-binding-templates/issues/176) and the corresponding [PR](https://github.com/w3c/wot-binding-templates/pull/183) where you can find some of the points that were discussed (if you want more details probably we should dig some minutes files from the calls). 

I remember that human readability was actually one of the pros of keeping the old design. We took that into account but, in the end, we preferred to follow [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986 ) more closely.

hadjian:
From the discussion I now  understand. Thx @relu91. We already implemented it in our connectivity solution.

> I remember a looong discussion about these topics. I've found an [issue](https://github.com/w3c/wot-binding-templates/issues/176) and the corresponding [PR](https://github.com/w3c/wot-binding-templates/pull/183) where you can find some of the points that were discussed (if you want more details probably we should dig some minutes files from the calls).
> 
> I remember that human readability was actually one of the pros of keeping the old design. We took that into account but, in the end, we preferred to follow [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986) more closely.

PR #309, opened 2023-10-16
Reorganizing uri variables section

### LINKED ISSUE #302: URI Variable for BACnet
There are a couple of unclear points regarding the use of URI variables in BACnet binding. I  think they are all the same issue but let me write them separately for now.

- In URI Variables, we see covIncrement (https://w3c.github.io/wot-binding-templates/bindings/protocols/bacnet/#uri-variables-0) but [Example 3](https://w3c.github.io/wot-binding-templates/bindings/protocols/bacnet/#example-form-variables) uses it like a form term.
- [Example 2](https://w3c.github.io/wot-binding-templates/bindings/protocols/bacnet/#example-affordance-variables) has URI Variables but the href does not contain them. This can be flagged as an error during parsing.
- The covIncrement is used in two ways but I am not sure if that is desired. One as a value of `@type` and the other as a key (`"bacv:covIncrement": 0.2`).

The whole problem seems to originate from the fact that we want to respect the BACnet uri scheme but want to add some input parameters to reading and observing. I think those should be all form terms and the URI Variables section can be removed.

egekorkan:
fixes #302 (the discussion is documented there)

### DIFF
--- a/bindings/protocols/bacnet/index.html
+++ b/bindings/protocols/bacnet/index.html
@@ -309,69 +309,6 @@ lowercase = %x61-7A  ; "a" to "z"
             This vocabulary is fully defined in the BACnet ontology published with this document.
         </p>
         <section>
-            <h3>URI Variables</h3>
-            <table class="def">
-                <thead>
-                    <tr>
-                        <th>Vocabulary term</th>
-                        <th>Description</th>
-                        <th>Assignment</th>
-                        <th>Type</th>
-                    </tr>
-                </thead>
-                <tbody>
-                    <tr>
-                        <td><code>CommandPriority</code></td>
-                        <td>Sets CmdPrio for Writes</td>
-                        <td>optional</td>
-                        <td>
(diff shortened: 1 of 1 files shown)


egekorkan:
Call of 18.10:
- We can use the forms for anything protocol-specific and in TD 2.0 we can have a nicer solution
- `default` can be used to indicate a constant value
- TBC

egekorkan:
Call of 25.10:
- @lu-zero : put the vocabulary terms involving the binding protocol only in the form, so consumers unaware of the binding can still safely consume the affordance as long there are forms it can consume
- @danielpeintner: We are creating a difference for the programmers (looking at affordance for uriVariables, looking at forms for bacnet). Agree with the reasoning but the solution is still not perfect. 
- @mjkoster : We should keep the separation of concerns and formalize this decision.

fennibay (commented):
I added my main comment on #302. We should finish that discussion first.

Also here I have a minor finding.

egekorkan:
As a result of https://github.com/w3c/wot-binding-templates/issues/302#issuecomment-1812825940, we will close this PR without merging. A follow up PR should change the URI Scheme to add the two parameters but also a warning saying "TD.next will change this"

PR #311, opened 2023-10-19
Replace to 'Custom devices' in Figure 1

### LINKED ISSUE #300: "Philips Hue with HTTP" in Figure 1 of the note of the WoT Binding Templates
There is the description of "Philips Hue with HTTP" in[ Figure 1](https://www.w3.org/TR/2023/NOTE-wot-binding-templates-20230928/#fig-building-block) of the note of the WoT Binding Templates. "Philips Hue" is the products of Philips. So I don't think its is appropriate to use "Philips Hue".

Thank you,

k-toumura:
Fix #300 

Replace "Philips Hue with HTTP" to "Custom devices with HTTP".

### DIFF
--- a/images/binding-templates.svg
+++ b/images/binding-templates.svg
@@ -6,7 +6,10 @@
    version="1.1"
    id="svg159"
    sodipodi:docname="binding-templates.svg"
-   inkscape:version="1.1.1 (3bf5ae0d25, 2021-09-20)"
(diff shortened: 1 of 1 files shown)


egekorkan (approved):
Thank you @k-toumura !

egekorkan:
Call of 25.10: We will merge this. Also see https://github.com/w3c/wot-architecture/pull/937

PR #308, opened 2023-10-16
Adding BACnet JSON Schema

egekorkan:
fixes #306 

I have also noticed some small things to fix in the index.html. The data mapping concept is rather new and I am not sure how to use them in examples.

PR #312, opened 2023-10-27
improve BACnet JSON schema

fennibay:
Match the schema to the domains and ranges of the fields.

This introduces nesting in the schema, which reflects the data model of BACnet.

### DIFF
--- a/bindings/protocols/bacnet/bacnet.schema.json
+++ b/bindings/protocols/bacnet/bacnet.schema.json
@@ -18,58 +18,235 @@
                         "SubscribeCOVproperty"
                     ]
                 },
-                "bacv:isISO8601": {
-                    "type": "boolean"
+                "bacv:hasDataType": {
+                    "$ref": "#/definitions/bacnetDataType"
                 },
-                "bacv:hasBinaryRepresentation": {
-                    "type": "string",
-                    "enum": [
-                        "hex",
-                        "dotted-decimal",
-                        "base64"
-                    ]
+                "bacv:covIncrement": {
+                    "type": "number",
+                    "minimum": 0
                 },
-                "bacv:hasMember": {
-                    "type": "string",
-                    "enum": [
-                        "hex"
-                    ]
-                },
-                "bacv:hasNamedMember": {
-                    "type": "string",
-                    "enum": [
-                        "hex"
-                    ]
-                },
-                "bacv:hasFieldName": {
-                    "type": "boolean"
-                },
-                "bacv:hasContextTags": {
-                    "type": "boolean"
-                },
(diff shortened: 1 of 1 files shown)


egekorkan:
So this is something we have agreed to not do (see links below) since it gives the wrong mapping. We want to say that if you are using a certain BACnet data, here is how you represent it in TDs. This would nudge people away from JSON Schema.

Links:
- https://github.com/w3c/wot-binding-templates/pull/308#issuecomment-1768676190
- https://github.com/w3c/wot-binding-templates/pull/308/commits/097d4be0f100669629e2fe7b819cb670391c09f8

fennibay:
Thanks @egekorkan for the info. What was the reasoning for removing the BACnet data types?

I'm fully with you that JSON schema should be the main contract. This PR proposes just a section under BACnet-specific forms, so that BACnet's complexity can be simplified to JSON. If this is not there, most types won't be mappable to JSON, or people are forced to use additional BACnet-specific metadata.

Some examples:
- In JSON we don't distinguish between 32-bit and 64-bit floats. Without the BACnet data type hint, a valid WriteProperty request cannot be prepared by the client
- In JSON, all objects have explicit fields. In BACnet they are sometimes implicitly tagged. Without BACnet data type the data cannot be prepared or parsed.
- With data-type mappings we can simplify BACnet-specific date, time-related formats into more web-like representations.

So, essentially I see these data types not making people farther from JSON, but rather bring them to the JSON/Web-world.

Let's discuss.

fennibay:
Short update after discussion with @egekorkan:
- The intention with this PR is _not_ to use BACnet types in the logical data schema.
- The intention is to model BACnet types under the Form, so the underlying BACnet implementation can handle the more complicated data model of BACnet.
- The logical data schema remains with JSON types.
- For example a BACnet unsigned value object will be a JSON integer with min: 0, and in BACnet form there will be a type hint that BACnet type is Unsigned and not Integer.

egekorkan (approved):
I had a chat with @fennibay and I understood. This goes in the direction that we talked about in the TD call. This maps the types in the affordance level to the concrete types that a bacnet driver requires.

@fennibay will provide an example in GitHub. A follow up PR will add a section between section 4 and 5 that explains the concept.

fennibay:
@egekorkan @ektrah I addressed the concerns in #314. I did not add a new section between the sections 4. URI Scheme and 5. BACnet Vocabulary, because the needed context for the payload examples is already under 6.2 WoT Data Schema and BACnet Types.

egekorkan:
Call of 08.11: We agree to merge. There has been commented that I will write in the #314

Issue #310, opened 2023-10-18
Data Mappings section has unclear purpose


egekorkan:
The section at https://w3c.github.io/wot-binding-templates/bindings/protocols/bacnet/index.html#datatype-mappings needs an introduction why this is needed.

egekorkan:
Other comment from @relu91 : The word "mapping" is confusing. We can add something like "How to use WoT Data Schema with BACnet types"

fennibay:
@egekorkan @relu91 I tried to address the concerns in #314, please check.

egekorkan:
We merged #312 and #314 but we need to address the following before closing this issue:

> We should be more assertive and say that the data types in forms should fit the data schema in the affordance level.

PR #314, opened 2023-11-06
Improve BACnet data mapping section

fennibay:
Data Mapping Section:
- Rename title (#310)
- Explain the purpose of the section (#310)
- Add BACnet payload examples to the data mapping section. 

Examples:
- Correct them inline with the updated JSON schema (#312)

Note: This already includes #312 and #313, so we can first merge those, or just merge this one instead.

### DIFF
--- a/bindings/protocols/bacnet/bacnet.schema.json
+++ b/bindings/protocols/bacnet/bacnet.schema.json
@@ -18,58 +18,235 @@
                         "SubscribeCOVproperty"
                     ]
                 },
-                "bacv:isISO8601": {
-                    "type": "boolean"
+                "bacv:hasDataType": {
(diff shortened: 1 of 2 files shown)


egekorkan (approved):
I suggest people to look at the example to really understand the design that is being proposed. I think this is fine for now, in future we can add a way to add descriptions to enum values in data schema.

egekorkan:
Call of 08.11:
- We are to merge but there has been some comments:
  - We should be more assertive and say that the data types in forms should fit the data schema in the affordance level.
  - This will have changes since we need to redesign the data mapping concept in td.next. This, together with endianness and wordswap of modbus are good use cases to drive that feature.

PR #318, opened 2023-11-17
BACnet protocol binding improvements

fennibay:
- Fix consistency issues between html, json schema and ttl
- Make CommandPriority and CovIncrement complete URI variables
- Remove covPeriod: This cannot be influenced from TD at all
- Add remark on expected changes in TD.next (ref #302)

I also took over some changes from #309.

### DIFF
--- a/bindings/protocols/bacnet/bacnet-model.ttl
+++ b/bindings/protocols/bacnet/bacnet-model.ttl
@@ -75,12 +75,6 @@ bacnet:covIncrement rdf:type owl:DatatypeProperty ;
                     rdfs:comment "Specifies the minimum change in that will cause a COV notification to be issued to subscribers, if missing the default value on the object will be used instead."@en ;
                     rdfs:seeAlso "ASHRAE 135-2020 BACnet 13.1 Change of Value Reporting" .
 
-bacnet:covPeriod rdf:type owl:DatatypeProperty ;
-                 rdfs:domain hctl:Form ;
-                 rdfs:range xsd:nonNegativeInteger ;
-                 rdfs:comment "Indicates the number of seconds to issue a COV notification periodically, even if there was no change. If 0 or missing, periodic COV notifications should be deactivated."@en ;
-                 rdfs:seeAlso "ASHRAE 135-2020 BACnet 13.1 Change of Value Reporting" .
-
 #################################################################
 #    Type information properties
 # BACnet has an own data model, which cannot always be derived
--- a/bindings/protocols/bacnet/bacnet-model.ttl
+++ b/bindings/protocols/bacnet/bacnet-model.ttl
@@ -64,11 +64,11 @@ Default BACnet service by td:OperationType:
 TODO: Can we specify the defaults in a machine-readable, RDF-standard way?
 
 Not mapped yet:
-- td:invokeAction, td:queryAction, td:cancelaction
+- td:invokeAction, td:queryAction, td:cancelAction
 - td:subscribeEvent, td:unsubscribeEvent
 - td:readAllProperties, td:writeAllProperties, td:readMultipleProperties, td:writeMultipleProperties, td:observeAllProperties, td:unobserveAllProperties
 - td:queryAllActions
-- td:subscribeallevents, td:unsubscribeallevents
+- td:subscribeAllEvents, td:unsubscribeAllEvents
 """@en .
 
 bacnet:CovIncrement rdf:type owl:Class ;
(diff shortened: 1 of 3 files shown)


bindings/protocols/bacnet/index.html
```diff
@@ -998,41 +1006,16 @@ <h2>Examples</h2>
...
-    "properties": {
-        "analog1": {
-            "type": "number",
-            "readOnly": true,
-            "forms": [{
-                "op": [ "observeproperty" ],
-                "href": "bacnet://5/0,1/85",
-                "bacv:usesService": "subscribeCOVproperty",
-                "bacv:covIncrement": 0.2,
+                "href": "bacnet://5/0,1/85?covIncrement={observeIncrement}",
```
egekorkan, line 1017: It should actually look a bit different I am not wrong. Following https://www.w3.org/TR/wot-thing-description11/#example-32 , this kind of URL should look like `"bacnet://5/0,1/85/{?observeIncrement}"` . If you want to make sure we see `covIncrement` in the URI, we may need another approach.

mjkoster: Example 32 in the TD spec isn't definitive in terms of the format of the URI variable query string format. RFC 3986 should be the definitive reference. https://datatracker.ietf.org/doc/html/rfc3986#section-3.4 It only says a query string can be a sequence of pchar (character set) and some additional characters you can use as delimiters, etc.

It is typical to format query string components with <identifier>=<value>, to allow for multiple parameters to be set explicitly. In the absence of a format mechanism, <identifier>=<value> is probably a good default mechanism. The URI Template should provide a way to describe the expected format of the query string, but URI templates are not well understood.

The uriVariable section is one I think needs a lot of work in TD.next, in conjunction with conceptualizing the parameterization of Property affordances, in general.

bindings/protocols/bacnet/bacnet-model.ttl
```diff
@@ -53,33 +53,50 @@ Additional parameters can be specified via URI variable syntax, e.g. ?bacnetCmdP
...
 - td:writeProperty -> WriteProperty
 - td:observeProperty -> SubscribeCOV, subscribe to object if no property ID is present in target, otherwise subscribe to property
-- td:unobserveProperty -> UnsubscribeCOV
+- td:unobserveProperty -> SubscribeCOV (will do an unsubscribe instead)
 
 TODO: Can we specify the defaults in a machine-readable, RDF-standard way?
 
 Not mapped yet:
-- td:invokeAction
+- td:invokeAction, td:queryAction, td:cancelaction
```
mjkoster, line 67: should it be camel case, i.e. cancelAction?

I didn't know these were in the TD namespace yet

fennibay: I was not sure, just followed the camel case convention. 😄 

But I doublechecked now, these are defined like this in JSON-LD context <https://www.w3.org/2022/wot/td/v1.1> and in the TD ontology: <https://www.w3.org/2019/wot/td#invokeAction> So we should be good here.

fennibay: td:cancelAction should be camel case, I will correct

egekorkan: For the record: the TD instances have all lower case, ontology is camel case with the translation done in JSON-LD Context.

bindings/protocols/bacnet/index.html
egekorkan: TD Call of 22.11: My comment is addressed since I was referring to compact form without realizing. Proposal is the long syntax, which is also fine.

egekorkan:
TD Call 22.11: @fennibay will cleanup the URI Scheme and then we can merge the PR asynchronously. He will also remove the draft from the name

bindings/protocols/bacnet/bacnet-model.ttl
```diff
@@ -53,33 +53,50 @@ Additional parameters can be specified via URI variable syntax, e.g. ?bacnetCmdP
...
 - td:subscribeEvent, td:unsubscribeEvent
 - td:readAllProperties, td:writeAllProperties, td:readMultipleProperties, td:writeMultipleProperties, td:observeAllProperties, td:unobserveAllProperties
+- td:queryAllActions
+- td:subscribeallevents, td:unsubscribeallevents
```
danielpeintner, line 71: Do we want to use camelCase here as well?

egekorkan: Yes we should: https://www.w3.org/2019/wot/td#unsubscribeAllEvents

bindings/protocols/bacnet/index.html
fennibay: I updated Chapter 4 now.

bindings/protocols/bacnet/bacnet-model.ttl
fennibay: Thank you, fixed it now.

bindings/protocols/bacnet/bacnet-model.ttl
fennibay: I fixed them now.

egekorkan (approved):
Looks good now, thanks @fennibay . As agreed in the call last week, we will merge this asynchronously.

PR #321, opened 2023-11-30
Fix typos in CoAP

danielpeintner:


egekorkan (approved):
If @ektrah or @JKRhb also approves it, we can merge it by the rule we agreed on at https://github.com/w3c/wot/blob/main/policies/async-decision.md#editorial-non-normative-changes

JKRhb (approved):
Looks good, thank you, @danielpeintner! :)

PR #324, opened 2023-11-30
Adding Jan Romann to CoAP codeowners

egekorkan:
Since he is an editor of the document, he should be added to code owners file


JKRhb:
Thank you, @egekorkan! I think for this change to work, I would need to be granted write access to the repository, would that be possible?

PR #323, opened 2023-11-30
Fix some typos in Modbus

danielpeintner:
Note: there are many "informations" that should be "information"
like in "Data response contains the informations requested by the client to the server."

PR #329, opened 2023-11-30
Modbus introduction improvement

### LINKED ISSUE #319: Better Explaining the current use case of the Modbus Binding
Ideally, the current Modbus binding template should have been written after a use case but we did not have such a process before. So, we should simply explain the use case in mind in the introduction section in a more clear fashion. To be more specific, this binding is meant for Modbus TCP devices (only mentioned in the URI Scheme section) where they mostly exhibit (writable) properties that have a simple request-response mechanism for interacting with them. There, we should explain that we talk about words, registers, and coils in Modbus, which would give way to the explanation of the word ordering issue we have been discussing in #293.

egekorkan:
fixes #319 

Will wait for #323


bindings/protocols/modbus/index.html
```diff
@@ -75,23 +75,33 @@
...
+            However, when Internet access is needed, it is encapsulated in TCP (called Modbus TCP) with Ethernet as the physical layer. 
+            Thanks to this encapsulation strategy, the Modbus protocol can reach remote nodes deployed in distant facilities over the Internet. 
+            Moreover, over the years, due to its simplicity and cost-effectiveness the Modbus protocol became a standard all over the world. 
+            This popularity, together with the advancement of the microcontroller/microprocessor, has led to a shift on how applications pack the information.
+            Building on this, this document we are focusing on the Modbus TCP.
+            Other versions of Modbus can be incorporated into this document in the future.
+        </p>
+        <p>
+            Concretely, this document describes how WoT TDs can be used to describe devices that use the Modbus TCP protocol.
+            In particular, the document explain how to create valid URLs and Forms for the different operations that the Modbus TCP protocol can perform.
```
danielpeintner, line 102:
```suggestion
            In particular, the document explains how to create valid URLs and Forms for the different operations that the Modbus TCP protocol can perform. 
```

bindings/protocols/modbus/index.html
```diff
@@ -75,23 +75,33 @@
...
-            encapsulated in a TCP/IP protocol and with Ethernet as a physical layer. Thanks to this encapsulation strategy, the Modbus protocol
-            can reach remote nodes deployed in distant facilities over the internet. Moreover, in the years due to its simplicity and cost-effectiveness 
-            the Modbus protocol becomes a standard all over the world. This fame together with the advancement of the microcontroller/microprocessor led to 
-            a shift on how applications pack the information. Today it is usual to store and read any type of data in the Holdings Registers like bit, bytes, words, float etc.
         </p>
         <p>
-            This document describes how the Web of Things specification can be use to present devices that use the Modbus protocol in a <a href="https://w3c.github.io/wot-architecture/#sec-thing-description">Thing Description</a>. 
-            In particular, the document explain how to create valid URLs and Forms for the different operations that the Modbus protocol can perform. 
+            All Modbus devices operate by exposing a certain set of their coils (bits) and registers (group of 16 bits) over the network.
```
relu91, line 84:
```suggestion
            All Modbus devices operate by exposing a certain set of coils (bits) and registers (group of 16 bits) over the network.
```
I think it is cleaner without their (?)

bindings/protocols/modbus/index.html
```diff
@@ -75,23 +75,33 @@
...
+            These coils or registers can be read or written to with a simple request-response mechanism.
```
relu91, line 85:
```suggestion
            These coils or registers can be read or written with a simple request-response mechanism.
```

bindings/protocols/modbus/index.html
```diff
@@ -75,23 +75,33 @@
...
+            together to represent a 32 bit integer.
+            Thus, a Modbus client can read a set of registers at once and obtain a large value.
+        </p>
+        <p>
+            The physical layer can be RS485 differential bus which has less susceptibility to the EMC interference.
+            This limits the usability of the protocol in short distance networks, typically within a kilometer.
+            However, when Internet access is needed, it is encapsulated in TCP (called Modbus TCP) with Ethernet as the physical layer. 
+            Thanks to this encapsulation strategy, the Modbus protocol can reach remote nodes deployed in distant facilities over the Internet. 
+            Moreover, over the years, due to its simplicity and cost-effectiveness the Modbus protocol became a standard all over the world. 
+            This popularity, together with the advancement of the microcontroller/microprocessor, has led to a shift on how applications pack the information.
```
relu91, line 96: This sentence feels a little bit out of the blue, I think removing 
> Today it is usual to store and read any type of data in the Holdings Registers like bit, bytes, words, float etc.

Decrease clarity.

bindings/protocols/modbus/index.html
```diff
@@ -75,23 +75,33 @@
...
+            Building on this, this document we are focusing on the Modbus TCP.
+            Other versions of Modbus can be incorporated into this document in the future.
+        </p>
+        <p>
+            Concretely, this document describes how WoT TDs can be used to describe devices that use the Modbus TCP protocol.
```
relu91, line 101:
```suggestion
        </p>
        <p>
            Concretely, this document describes how WoT TDs can be used to describe devices that use the Modbus TCP protocol.  Other versions of Modbus can be incorporated into this document in the future.
```

relu91 (changes requested):
I left some suggested changes below, but I noticed too late that they were on the main index.html document instead of index.template.html I hope they can be incorporated anyway without too much of effort.

bindings/protocols/modbus/index.html
egekorkan: From a small reading, both formulations are correct but I will fix accordingly since without the "to" seems more intuitive.

bindings/protocols/modbus/index.html
egekorkan: This was removed actually, I don't see it?

egekorkan:
@relu91 I have addressed your feedback (except for one comment). Feel free to resolve comments.

egekorkan:
Call of 13.12: @relu91 will have a final look on the comment above and will merge asynchronously if it is fine.

bindings/protocols/modbus/index.html
egekorkan: This was about reintroducing that sentence. I have done that in the latest commit. Resolving.

relu91 (approved):
Good to go now (there are some changes in the ontology HTML that I don't understand but since it is generated we can go!)

egekorkan:
> Good to go now (there are some changes in the ontology HTML that I don't understand but since it is generated we can go!)

Me neither but they are only whitespace changes. Merging :)

PR #336, opened 2023-12-05
Modbus Dataproperty terms changed to align with best practices

### LINKED ISSUE #335: modbus ontology
The Modbus ontology Data property names needs to be updated to align with common naming best practices.

mahdanoura:
fixes #335


bindings/protocols/modbus/index.html
egekorkan, line 1: Are the changes to remove those terms intentional?

bindings/protocols/modbus/ontology.ttl
```diff
@@ -129,54 +129,54 @@ binding:assignment rdf:type owl:AnnotationProperty .
 #################################################################
 
 ###  https://www.w3.org/2019/wot/modbus#hasAddress
-:hasAddress rdf:type owl:DatatypeProperty ;
+:address rdf:type owl:DatatypeProperty ;
```
sebastiankb, line 132: my expectation would be that hasAddress, hasCodeNumber, etc are correct. In other WoT ontologies, we use the same pattern (e.g., https://www.w3.org/2019/wot/hypermedia#hasTarget).

The context file is the place, which should do the mapping from address to hasAddress, code to hasCode, etc and vice versa.

bindings/protocols/modbus/index.html
mahdanoura: I didn't remove any term, just updated the names to not include the "has", but maybe the tool does additional magic

mahdanoura:
> I'm ok with changing (even though funny enough I thought that convention was to use always a verb, but maybe it is just for ObjectProperties?). Just remember that the `context.json` should be updated.

It is definitly not incorrect to have them, we can still stay with the current convention. As @sebastiankb also pointed they are used in the other WoT ontologies in a similar manner. However, when constructing RDF statements they should be also understandable in the meaning. The object properties have Verbs. They should read like triple statements, e.g., 
[resource] John Doe has a [property] address which is [value] X.

egekorkan:
TD Call 13.12:
- We are inconsistent in general so we should do this in one go for all ontologies. This should be a guideline for ontology editors
- Since this change impacts the URLs (e.g. `https://www.w3.org/2019/wot/modbus#hasAddress` becoming `https://www.w3.org/2019/wot/modbus#address`), implementations who use the ontology will be affected. We will get more input for this from known users, such as from @sebastiankb .

Issue #337, opened 2023-12-06
[Modbus] there are many "informations" that should be "information"


danielpeintner:
> Note: there are many "informations" that should be "information"

Question: Do you want to deal with it separately ?
I can provide yet another PR to show what I mean if you like...

_Originally posted by @danielpeintner in https://github.com/w3c/wot-binding-templates/issues/323#issuecomment-1833399715_

PR #343, opened 2023-12-13
[Modbus] Typos and slight rewordings

danielpeintner:
fixes https://github.com/w3c/wot-binding-templates/issues/337


egekorkan (approved):
I think there are no conflicts with https://github.com/w3c/wot-binding-templates/pull/329 so we can merge it in parallel

egekorkan (approved):
Also approved by @relu91 in the TD call. Merging

PR #352, opened 2024-02-19
Min Polling Interval in CoAP

### LINKED ISSUE #349: Max Polling Time in CoAP
Similar to Modbus, some CoAP devices also have a maximum polling time they support due to their resources. This means that the consumers are advised/recommended not to poll more frequently. 

The devices in question are using TD 1.1 so it can only be in the CoAP binding. Once the use cases start, I will propose this as a new term for TD.

ektrah:
Resolves #349


egekorkan (changes requested):
I agree with the content however we are using the spec generation tooling. So the table row should be contributed at https://github.com/w3c/wot-binding-templates/blob/main/bindings/protocols/coap/ontology.ttl#L34 and the text should be done to index.template.html in the same directory. We also need to update the schema but that can be done easily.

bindings/protocols/coap/index.html
```diff
@@ -499,6 +540,16 @@ <h3>Form terms</h3>
...
                         </td>
                     </tr>
+                    <tr>
+                        <td><code>cov:minPollingInterval</code></td>
+                        <td>
+                            Provides a hint of what minimum polling interval (in seconds) the server is willing to accept before rejecting the client with a 4.29 (Too Many Requests) response [[RFC8516]].
+                        </td>
+                        <td>optional</td>
+                        <td>
+                            <a href="http://www.w3.org/2001/XMLSchema#unsignedInt">unsignedInt</a>
```
danielpeintner, line 550: No objection. I just want to mention that the possible range is rather large. Maybe `unsignedShort` is sufficient but I don't know whether this really matters in this case..

ektrah: According to [RFC 8516](https://www.rfc-editor.org/rfc/rfc8516.html#section-3), which defines the 4.29 (Too Many Requests) response code, "the Max-Age option is used to indicate the number of seconds after which the server assumes it is OK for the client to retry the request". And, according to [RFC 7252](https://datatracker.ietf.org/doc/html/rfc7252#section-5.10.5), "the option value is an integer number of seconds between 0 and 2**32-1 inclusive (about 136.1 years)."

So, yes, the range is rather large. But that's how CoAP defines it. Since we want to preempt the server's 4.29 error response by specifying the interval in the TD, I think we should go for the same range here.

egekorkan:
In the call of 22.02, we have agreed on the content change. @egekorkan will fix the tooling to use the turtle file in the generation of the index.html

egekorkan:
I have done the changes as agreed upon. Looking at https://github.com/w3c/wot-binding-templates/pull/352/commits/97156717b5528b39ade19360ee4ff167fca0c2ec there is no real change in the index.html. I have also added this field to json schema

egekorkan:
@ektrah and @JKRhb can you have a final look and give ok? We can merge it asynchronously :)

bindings/protocols/coap/index.html
```diff
@@ -282,7 +325,6 @@ <h3>Observing Resources</h3>
             </p>
             <p>
                 Support for observing resources can be indicating by including a <code>subprotocol</code> value of <code>cov:observe</code> in a form.
-                However, a consumer can always fall back to polling a resource if observing a resource is permanently or temporarily not possible.
```
ektrah, line 285: This line was removed in d271c16daa0d34a229762f0b72a80033cda665b2 because it is redundant with the following paragraph, but was added back in 97156717b5528b39ade19360ee4ff167fca0c2ec. I have removed it again in 3713e2d241b579d09ca99f5792c9bb2cdc67ea31.

index.html looks good to me. I trust that the other files have all the necessary changes. So 👍  from me.

egekorkan: Perfect, thanks!

JKRhb:
> @JKRhb if fine with you as well, you can merge it :)

Thank you, Ege, looks good to me! :)

PR #356, opened 2024-04-01
fix: move context tag from Sequence/Choice to NamedMember

fennibay:
This allows (1) partially tagged sequences and choices and (2) modelling the context tag of each member.

p.s. I don't know the reason of the change in modbus/ontology.html. Must be auto-formatting commit hook or stg similar.


egekorkan (approved):

egekorkan:
Modbus changes are white-space only but still curious. Maybe a previous commit did not format correctly?

egekorkan:
TD call May 2: approved by @mjkoster

PR #358, opened 2024-04-08
Small Modbus fixes

egekorkan:
There were some leftover commas at the end of forms and one form had leftover modv:address even though it is now in the URI


relu91 (approved):
Thanks, good to go from my side.

egekorkan:
Editor has approved an editorial change, which we can merge as per https://github.com/w3c/wot/blob/main/policies/async-decision.md

PR #359, opened 2024-04-09
Modbus fixes

egekorkan:
This PR fixes a couple of small but important aspects:

- Types: We have noticed that type information like https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#unsignedByte are not aligned with the binding template at even though our ontology was correct (https://w3c.github.io/wot-binding-templates/bindings/protocols/modbus/ontology.html#payloaddatatype). This is due to lowercasing in the template.sparql file . I had to fix the schema manually.
- Editor was missing in the ontology
- Schema was using wrong operation names where an `s` was missing. This is the perfect example of needing toolchain since this was an oversight from our side for a long time.


relu91 (approved):

bindings/protocols/modbus/index.html
egekorkan, line 1: lines 206 to 236 are not changes, the toolchain has reordered them

bindings/protocols/modbus/index.html
```diff
@@ -541,6 +529,18 @@ <h3>Default mappings</h3>
...
+            <td><code>writeallproperties</code></td>
+            <td><code>"modv:function": "writeMultipleHoldingRegisters"</code></td>
+        <tr>
+        
+
+        <tr>
+            <td><code>writemultipleproperties</code></td>
+            <td><code>"modv:function": "writeMultipleHoldingRegisters"</code></td>
+        <tr>
+
```
egekorkan, line 543: this is also reordering

bindings/protocols/modbus/ontology.ttl
egekorkan, line 1: please hide whitespace changes when reviewing this

bindings/protocols/modbus/index.html
```diff
@@ -203,37 +203,37 @@ <h3>Form terms</h3>
...
         
 
         <tr>
-            <td><code>modv:timeout</code></td>
-            <td>Modbus response maximum waiting time. Defines how much time the runtime should wait until it receives a reply from the device.</td>
+            <td><code>modv:mostSignificantWord</code></td>
+            <td>When <code>true</code>, it describes that the word order of the data in the Modbus message is the most significant word first (i.e., no word swapping). When <code>false</code>, it describes the least significant word first (i.e. word swapping)</td>
             <td>optional</td>
-            <td><a href="http://www.w3.org/2001/XMLSchema#integer">integer</a></td>
+            <td><a href="http://www.w3.org/2001/XMLSchema#boolean">boolean</a></td>
```
danielpeintner, line 236: Did `integer` disappear?

egekorkan: No this is the type of mostSignificantWord . integer is the type of timeout which got moved above

danielpeintner: Ahh, I see

egekorkan:
Call of 11.04: 
- @relu91 We have noticed that the function names need a second look since the name, label and descriptions sometimes have a mismatch. We should see the modbus spec as well (e.g. is it read coil or read coilS)
- In any case, we should make sure to review the table content much better since the table content is what is used in schemas or code, thus used by software that is very strict i.e. the presence of an s or capitalization means a totally different word.

egekorkan:
The labels and descriptions are fine for now and the values are copy-pasted into implementations anyways so no need to change with this PR. Merging as agreed upon

PR #375, opened 2024-07-05
Add dark mode support

JKRhb:
As also discussed over at https://github.com/w3c/wot-thing-description/pull/2029, this simple PR enables dark mode support for all the documents (or rather: their current draft versions) in this repository.

Note that there are some unrelated changes occurring with the Modbus ontology – maybe that one should be regenerated on the `main` branch?

***
<a href="https://pr-preview.s3.amazonaws.com/JKRhb/wot-binding-templates/pull/375.html" title="Last updated on Jul 10, 2024, 10:07 AM UTC (182868a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-binding-templates/375/18051f3...JKRhb:182868a.html" title="Last updated on Jul 10, 2024, 10:07 AM UTC (182868a)">Diff</a>

### DIFF
--- a/bindings/index.template.html
+++ b/bindings/index.template.html
@@ -3,6 +3,7 @@
 
 <head>
     <meta charset='utf-8'>
+    <meta name='color-scheme' content='light dark'>
     <script src='https://www.w3.org/Tools/respec/respec-w3c' class='remove'></script>
     <script class='remove'>
         var respecConfig = {
--- a/bindings/payloads/xml/index.html
+++ b/bindings/payloads/xml/index.html
@@ -2,6 +2,7 @@
 <html>
 <head>
     <meta charset='utf-8'>
+    <meta name='color-scheme' content='light dark'>
     <script src='https://www.w3.org/Tools/respec/respec-w3c' class='remove'></script>
     <script class='remove'>
         var respecConfig = {
--- a/bindings/platforms/hue/index.html
+++ b/bindings/platforms/hue/index.html
@@ -2,6 +2,7 @@
 <html>
 <head>
     <meta charset='utf-8'>
+    <meta name='color-scheme' content='light dark'>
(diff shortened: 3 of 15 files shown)


egekorkan (approved):
Hiding white spaces is recommended for reviewing the modbus ontology changes. I would not complicate the process by requiring another PR

Also, since these are simple changes that have "nothing" to do with the individual bindings, we can skip the required reviews by binding authors. 

Given an editorial change and approved by one editor, this PR can be merged async after one week if no diverging opinions arise and will not be discussed in the call to save time. See https://github.com/w3c/wot/blob/main/policies/async-decision.md

bindings/index.template.html
```diff
@@ -3,6 +3,7 @@
 
 <head>
     <meta charset='utf-8'>
+    <meta name="color-scheme" content="light dark" />
```
danielpeintner, line 6:
```suggestion
    <meta name="color-scheme" content="light dark">
```

Nit: Just because of consistency I think we should use the `meta` term without the ending `/>` as we have the line above. Personally I use closing tags always but since the old code didn't ...

The same would apply for the quotes though also... `"` vs `'`

JKRhb: Thanks, I applied that via 4fd2a55bd613839e27441639a23af86c2784e419 :)

JKRhb: I also adjusted it in other places now as well. I guess as soon as we have some kind of linting/auto-formatting step here as well, most of these consistency questions will be resolved for good :)

relu91 (approved):
I agree with Ege we can go ahead with this PR! On the other hand I'm wondering why this white space problem is still happening 🙄

JKRhb:
Thanks again for your feedback everyone, I took the freedom of squashing the `fixup!` commits so that this PR should now be ready to be merged once one full week has passed.

egekorkan:
No further comments, merging.

PR #378, opened 2024-07-25
Registry Requirements Update

egekorkan:
To prepare for today's discussion, I have done some simple changes and I am noting them as comments on GitHub.

Additionally, I have changed some "TODO"s to "DISCUSS". TODOs will remain points that are not clear yet due to spec not being ready (e.g. data mapping). The other points are open for discussion.


registry-requirements.md
```diff
@@ -23,7 +23,6 @@ It is unrealistic to incorporate a complete list of bindings into a REC document
 To choose a mechanism and the rules, we document our expectations below:
 
 - A binding SHOULD be written by people with a good understanding of the protocol, media type (or similar), and not necessarily the TD Editors. This includes people and organizations inside and outside of the WoT WG.
-  - TODO: Ask existing registry "managers" about their opinion on this proposal. This will be discussed in the breakout on March 12.
```
egekorkan, line 26: This has been already discussed

registry-requirements.md
```diff
@@ -44,24 +43,22 @@ To choose a mechanism and the rules, we document our expectations below:
...
+  - DISCUSS: Clarify what happens when two ecosystems like OCF and LwM2M both use CoAP binding. The initial thinking is to register them as separate entries but clarify what they use from CoAP binding. A layered registry can be thought of as similar to language tags with `en` extending to `en-us` and `en-uk`, where the tags and entries are different but the association with `en` is expressed in the id. In our case, this would be `coap-ocf` and `coap-lwm2m`. A higher-level binding SHOULD NOT override or conflict with a lower-level binding, and this should be verified by the custodian, e.g., `cov:method` in CoAP binding should not be turned into `cov:operation` in the higher-level binding. The namespace (prefix and its values) defined in a binding CANNOT be redefined in any other binding.
+- In a TD, a binding SHOULD be identifiable by the terms in a form such as `href`, `contentType`, `subprotocol`, or connection information found in the root of the TD.
```
egekorkan, line 47: I hope to have consensus on this formulation

registry-requirements.md
```diff
@@ -44,24 +43,22 @@ To choose a mechanism and the rules, we document our expectations below:
...
+  - DISCUSS: How to identify a higher-level binding in the form
 - A binding that uses a protocol MUST map at least one WoT operation (`op` keyword values such as `readproperty`) to a protocol message and vice versa
   - Reasons Behind the Requirement:
     - Otherwise, it is not binding the protocol to WoT and cannot be useful.
 - A binding that uses a serialization format via the `contentType` keyword, MUST mention how the Data Schema terms should be used to describe the messages.
   - Reasons Behind the Requirement:
     - Avoid submission of a binding like "XML Binding" that only says "Use `contentType:application/xml` and nothing more. That alone would not be enough to serialize correct messages based on the data schema.
   - TODO: We will need additional mechanisms (including vocabulary terms) to ensure that it is possible to use other media types.
-- TODO: Explain that there are no categories
-- TODO: Analyze <https://github.com/w3c/ttwg/tree/main/boilerplate/registry>
```
egekorkan, line 64: These two points are already discussed. We agreed that there will be no categories and the analysis of TTWG boilerplate is done already.

registry-requirements.md
```diff
@@ -13,86 +12,34 @@ From the charter:
...
-- Differentiating entry into the registry and update
-- Technical submission mechanism. How does a binding get submitted? Is it a PR, an issue linking to an existing document, or an email? See the submission mechanism fields above.
-- Whether we should have one or multiple registries
+This is what is put into the table and not what the binding should contain.
+At each submission, these information must be present in the submission.
 
+- Name of the binding
+  - Examples: `HTTP Binding Template`, `CoAP Binding Template`, 
+- Link to the binding: Stable and dated link
+  - Examples: `https://w3c.github.io/wot-binding-templates/bindings/protocols/http/index.html`
```
TallTed, line 32: This example URL should be dated.

registry-requirements.md
```diff
@@ -13,86 +12,34 @@ From the charter:
...
+- Binding ontology prefix
+  - Examples: `modv`, `htv`, `cov`
+- Binding Identification in TD: URI Scheme or other TD terms that are reserved for this binding.
+  - Examples: `subprotocol:sse`, `href:"http://example.com"
```
TallTed, line 36:
```suggestion
  - Examples: `subprotocol:sse`, `href:"http://example.com"`
```

registry-requirements.md
```diff
@@ -13,86 +12,34 @@ From the charter:
...
-A preliminary list of rules (needs more iteration):
+This is what the binding has to contain in order to go into the table
 
-- Entry format (i.e. what is put into the TD document and not what the linked document should contain)
-  - Name of the binding
-  - Link to the binding: Stable link
-  - (possibly) The binding prefix
-- Requirements on the submitted document:
-  - To be clarified but the initial list for protocols at <https://w3c.github.io/wot-binding-templates/#creating-a-new-protocol-binding-template-subspecification> and <https://w3c.github.io/wot-binding-templates/#protocol-bindings-table>
+- To be clarified but the initial list for protocols at <https://w3c.github.io/wot-binding-templates/#creating-a-new-protocol-binding-template-subspecification> and <https://w3c.github.io/wot-binding-templates/#protocol-bindings-table>
```
TallTed, line 42:
```suggestion
- To be clarified, but the initial list for protocols is at <https://w3c.github.io/wot-binding-templates/#creating-a-new-protocol-binding-template-subspecification> and <https://w3c.github.io/wot-binding-templates/#protocol-bindings-table>
```

registry-requirements.md
```diff
@@ -122,3 +69,48 @@ Same content as existing binding
 (Think about how to manage the set of semantic artifacts/ontologies)
 
 Create a repository template to make it easy for people to create new bindings
+
+## Entry Requirements
+
+
+## Basic Requirements
```
TallTed, line 76: Appears to be meant to be a subsection of `Entry Requirements`
```suggestion
### Basic Requirements
```

registry-requirements.md
```diff
@@ -122,3 +69,48 @@ Same content as existing binding
...
+
+To choose a mechanism and the rules, we document our expectations below:
+
+- A binding SHOULD be written by people with a good understanding of the protocol, media type (or similar), and not necessarily the TD Editors. This includes people and organizations inside and outside of the WoT WG.
```
TallTed, line 80:
```suggestion
- A binding SHOULD be written by people with a good understanding of the protocol and media type (or similar), who are not necessarily the TD Editors. This includes people and organizations inside and outside of the WoT WG.
```

registry-requirements.md
```diff
@@ -122,3 +69,48 @@ Same content as existing binding
...
+- The binding registry MUST be a separate document but associated with a TD version.
+  - Reasons Behind the Requirement: It is easier to update in the long term.
+- Association of a binding with the TD specification (registry entry) SHOULD be confirmed by the WoT Working Group. In other words, a person needs some permission and/or confirmation to authoritatively say that a given binding can be used with TD version X. The custodian of this registry is the WoT WG.
+  - Reasons Behind the Requirement: WoT WG is the manager of the registry.
+- If the WoT WG no longer exists, the W3C Team or their delegated entity becomes the custodian.
+  - Reasons Behind the Requirement: It should be possible to maintain the registry without the WoT WG.
+- It SHOULD be possible to register, update, or remove a registry entry for associating a binding to a specific TD specification outside of the REC lifecycle of the given TD specification.
```
TallTed, line 92:
```suggestion
- It SHOULD be possible to register, update, or remove a registry entry that associates a binding to a specific TD specification outside of the REC lifecycle of the given TD specification.
```

registry-requirements.md
```diff
@@ -122,3 +69,48 @@ Same content as existing binding
...
+  - Reasons Behind the Requirement:
+    - Instead of WG learning each new protocol and media type, it is more efficient for people with a good understanding of the protocol or media type to write a binding.
+    - Engaging other communities.
+- A binding SHOULD correspond to specific TD specification version(s).
+  - Reason Behind the Requirement: A binding may not fit newer or older versions of a TD specification (e.g., `readproperty` can become `readprop`, or a new operation can arrive). Thus, at the time of writing a binding, it needs to be associated with one or more known TD specification versions.
+- The binding registry MUST be a separate document but associated with a TD version.
+  - Reasons Behind the Requirement: It is easier to update in the long term.
+- Association of a binding with the TD specification (registry entry) SHOULD be confirmed by the WoT Working Group. In other words, a person needs some permission and/or confirmation to authoritatively say that a given binding can be used with TD version X. The custodian of this registry is the WoT WG.
```
TallTed, line 88: It would be better to make a registry entry require that some test(suite) be passed, making the requirement clearly technical. As written, politics could come into play, to the detriment of all involved.

registry-requirements.md
```diff
@@ -122,3 +69,48 @@ Same content as existing binding
...
+  - Reasons Behind the Requirement: It should be possible to maintain the registry without the WoT WG.
+- It SHOULD be possible to register, update, or remove a registry entry for associating a binding to a specific TD specification outside of the REC lifecycle of the given TD specification.
+  - Reasons Behind the Requirement:
+    - Allowing new protocols to be bound to any TD version. 
+    - The process should outlive the WG if the WG ever comes to stop functioning or if the WG decides to not work on a TD specification version anymore.
+- There SHOULD NOT be two bindings for the same protocol, media type etc.
+  - Reasons Behind the Requirement:
+    - It breaks interoperability otherwise
+    - We should avoid confusing users, i.e. which binding should I use?
+    - It would complicate implementations
```
TallTed, line 100:
```suggestion
    - The process should outlive the WG, in case the WG ever dissolves or the WG decides to stop maintaining a given TD specification version.
- There SHOULD NOT be two bindings for the same protocol, media type, etc.
  - Reasons Behind the Requirement:
    - Such multiple bindings would break interoperability
    - Such multiple bindings would confuse users, i.e., which binding should I use?
    - Such multiple bindings would complicate implementations
```

registry-requirements.md
```diff
@@ -122,3 +69,48 @@ Same content as existing binding
...
+  - DISCUSS: Clarify what happens when two ecosystems like OCF and LwM2M both use CoAP binding. The initial thinking is to register them as separate entries but clarify what they use from CoAP binding. A layered registry can be thought of as similar to language tags with `en` extending to `en-us` and `en-uk`, where the tags and entries are different but the association with `en` is expressed in the id. In our case, this would be `coap-ocf` and `coap-lwm2m`. A higher-level binding SHOULD NOT override or conflict with a lower-level binding, and this should be verified by the custodian, e.g., `cov:method` in CoAP binding should not be turned into `cov:operation` in the higher-level binding. The namespace (prefix and its values) defined in a binding CANNOT be redefined in any other binding.
```
TallTed, line 101:
```suggestion
  - DISCUSS: Clarify what happens when two ecosystems like OCF and LwM2M both use CoAP binding. The initial thinking is to register them as separate entries and clarify what they use from CoAP binding. A layered registry can be thought of as similar to language tags with `en` extending to `en-us` and `en-uk`, where the tags and entries are different but the association with `en` is expressed in the ID. In our case, this would be `coap-ocf` and `coap-lwm2m`. A higher-level binding SHOULD NOT override or conflict with a lower-level binding, and this should be verified by the custodian, e.g., `cov:method` in CoAP binding should not be turned into `cov:operation` in the higher-level binding. The namespace (prefix and its values) defined in a binding CANNOT be redefined in any other binding.
```

registry-requirements.md
```diff
@@ -122,3 +69,48 @@ Same content as existing binding
...
+- In a TD, a binding SHOULD be identifiable by the terms in a form such as `href`, `contentType`, `subprotocol`, or connection information found in the root of the TD.
+  - Reasons Behind the Requirement:
+    - This avoids conflicts that are mentioned in the previous requirement
+  - TODO: These terms should be refined based on the additions/changes to the TD 2.0 mechanism. E.g. introducing a `protocol` term, or restrictions on URI scheme and `subprotocol` combination, data mapping etc.
```
TallTed, line 105:
```suggestion
  - TODO: These terms should be refined based on the additions/changes to the TD 2.0 mechanism, e.g., introducing a `protocol` term, or putting restrictions on URI scheme and `subprotocol` combination, data mapping, etc.
```

registry-requirements.md
```diff
@@ -122,3 +69,48 @@ Same content as existing binding
...
+  - DISCUSS: We need to clarify whether the URI scheme and the media type MUST be registered in IANA first.
+    - Provisional registration could reduce the overhead. Any new conflicting ones would bring up a discussion but it can still result in "our" provisional getting demoted.
+    - Pros: more stable. Cons: More overhead and work
+  - DISCUSS: How to identify a higher-level binding in the form
+- A binding that uses a protocol MUST map at least one WoT operation (`op` keyword values such as `readproperty`) to a protocol message and vice versa
```
TallTed, line 110:
```suggestion
- A binding that uses a protocol MUST map at least one WoT operation (`op` keyword value such as `readproperty`) to a protocol message and vice versa
```

registry-requirements.md
```diff
@@ -122,3 +69,48 @@ Same content as existing binding
...
+  - Reasons Behind the Requirement:
+    - Otherwise, it is not binding the protocol to WoT and cannot be useful.
```
TallTed, line 112:
```suggestion
    - Lacking this requirement, the protocol is not bound to WoT and cannot be useful.
```

registry-requirements.md
```diff
@@ -122,3 +69,48 @@ Same content as existing binding
...
+- A binding that uses a serialization format via the `contentType` keyword, MUST mention how the Data Schema terms should be used to describe the messages.
+  - Reasons Behind the Requirement:
+    - Avoid submission of a binding like "XML Binding" that only says "Use `contentType:application/xml` and nothing more. That alone would not be enough to serialize correct messages based on the data schema.
```
TallTed, line 115:
```suggestion
    - Avoid submission of a binding like "XML Binding" that says "Use `contentType:application/xml` and nothing more. That alone would not be enough to serialize correct messages based on the data schema.
```

TallTed (changes requested):

registry-requirements.md
egekorkan: We don't have dated versions of them yet :/ We can put a date of last update?

TallTed: Or use an `example` URL... I'm guessing at what such a dated URL might look like --
```suggestion
  - Examples: `https://w3c.example/wot-binding-templates/bindings/protocols/http/index-20240726.html`
```

registry-requirements.md
egekorkan: Not in this case. The entry requirements will be what is written in the registry section. The basic requirements are a big set of requirements and will be moved into entry requirements in some cases. There is still some work to do in this PR so I am putting it into the draft status.

registry-requirements.md
egekorkan:
```suggestion
  - Examples: `https://www.w3.org/TR/wot/binding-templates/http-20240726/index.html`
```

I am not sure but it should start at least like this. Given this is just an example at this point which gives both the stability and date properties, it should be fine.

registry-requirements.md
egekorkan: Added a DISCUSS comment about this

registry-requirements.md
```diff
@@ -44,24 +43,22 @@ To choose a mechanism and the rules, we document our expectations below:
...
+- In a TD, a binding SHOULD be identifiable by the terms in a form such as `href`, `contentType`, `subprotocol`, or connection information found in the root of the TD.
   - Reasons Behind the Requirement:
     - This avoids conflicts that are mentioned in the previous requirement
   - TODO: These terms should be refined based on the additions/changes to the TD 2.0 mechanism. E.g. introducing a `protocol` term, or restrictions on URI scheme and `subprotocol` combination, data mapping etc.
-  - TODO: We need to clarify whether the URI scheme and the media type MUST be registered in IANA first.
+  - DISCUSS: We need to clarify whether the URI scheme and the media type MUST be registered in IANA first.
```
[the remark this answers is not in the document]
egekorkan: @danielpeintner @lu-zero your comments are now part of the document

registry-requirements.md
```diff
@@ -24,16 +24,17 @@ A preliminary list of rules that is extending https://www.w3.org/2023/Process-20
 ### Entry format 
 
 This is what is put into the table and not what the binding should contain.
-At each submission, these information must be present in the submission.
+At each submission, these information MUST be present in the submission and MUST not conflict with existing bindings.
```
TallTed, line 27:
```suggestion
At each submission, this information MUST be present in the submission and MUST not conflict with existing bindings.
```

registry-requirements.md
```diff
@@ -52,26 +52,23 @@ Each entry MUST contain this information, and all parts of the entry MUST not co
...
 - Versioning of registry entries (see https://github.com/w3c/wot/tree/main/registry-analysis#versioning)
   - Ege: We do not allow updates of a registry document content. A new version of a binding is a resubmission and optional deprecation of the old one. Only new features need new implementations though so it is not a completely new registration.
   - DISCUSS: Do we allow two versions of a binding to stay in the registry?
-    - Ege: At least per TD version, it should be avoided but I think multiple initial entries should be allowed in case of diverging opinions to collect implementation feedback. If there is a drastically new binding of an existing one, two can be allowed to be kept but this should not be encouraged. This will causes issues with prefix, URI scheme based detection etc.
+    - Ege: At least per TD version, it should be avoided but I think multiple initial entries should be allowed in case of diverging opinions to collect implementation feedback. If there is a drastically new binding of an existing one, two can be allowed to be kept but this should not be encouraged. This will cause issues with prefix, URI scheme based detection, or other form elements etc.
```
TallTed, line 68:
```suggestion
    - Ege: At least per TD version, it should be avoided, but I think multiple initial entries should be allowed in case of diverging opinions, to collect implementation feedback. If there is a drastically new binding of an existing one, two can be allowed to be kept, but this should not be encouraged. This will cause issues with prefix, URI scheme-based detection, other form elements, etc.
```

registry-requirements.md
```diff
@@ -83,8 +80,9 @@ Each entry MUST contain this information, and all parts of the entry MUST not co
 
 - If the WoT WG no longer exists, the W3C Team or their delegated entity becomes the custodian.
   - Reasons Behind the Requirement: Maintaining the registry without the WoT WG should be possible.
+- DISCUSS: Who is the reviewer? Same as custodian? Look into https://datatracker.ietf.org/doc/html/rfc6838 . How is the "separation of concerns"?
```
TallTed, line 83:
```suggestion
- DISCUSS: Who is the reviewer? Same as custodian? Look into [RFC6838](https://datatracker.ietf.org/doc/html/rfc6838). How is the "separation of concerns"?
```

registry-requirements.md
```diff
@@ -70,17 +70,16 @@ Each entry MUST contain this information, and all parts of the entry MUST not co
       - Luca: Prefix versioning would require calling a new binding driver inside the implementation. Not using the prefix versioning implies JSON-LD processing, which not all implementations use.
     - We should provide an example and also look at RFC 6838
 - Deletion and deprecation (see https://github.com/w3c/wot/tree/main/registry-analysis#deletion-and-deprecation-of-registry-entries)
-  - Ege: No entry is ever deleted. Deprecated entries are moved to another table or to the bottom of the table and marked clearly deprecated and colored differently.
-
-- Status values: Initial (provisional, draft) -> Final -> Deprecated
-  - Merged initial entries trigger a "Call for Implementation"
+  - No entry is ever deleted. Deprecated entries are moved to another table or to the bottom of the table and marked clearly deprecated and colored differently.
```
TallTed, line 73:
```suggestion
  - No entry is ever deleted. Deprecated entries are moved to another table, or are clearly marked deprecated, colored differently, and moved to the bottom of the table.
```

TallTed (changes requested):

registry-requirements.md
```diff
@@ -42,7 +42,7 @@ Each entry MUST contain this information, and all parts of the entry MUST not co
 
 1. Name of the binding
   - Examples: `HTTP Binding Template`, `CoAP Binding Template`
-2. Link to the binding: Stable and dated link
+2. Link to the binding document: Stable link whose content cannot change (e.g. a date, version number etc)
```
TallTed, line 45:
```suggestion
2. Link to the binding document: Stable link whose content cannot change (e.g., a date, version number, etc.)
```

registry-requirements.md
```diff
@@ -94,9 +94,14 @@ What does the binding have to contain to go into the table
 3. DISCUSS: What is the objective mechanism to confirm the initial entry, i.e., status equals "initial"?
   - Ege: Initial entry (provisional, draft) is checked for document correctness, i.e., mapping at least one protocol, URI scheme, examples of each mapped operation, "casual" introduction, and abstract. However, we do not check if someone tries to map `readproperty` to a non-existent HTTP method.
   - Cris: Merging the initial entry would trigger a "Call for Implementation". Where discussions on implementation experience should be collected?
-  - DISCUSS: Should the binding document be required to follow W3C copyright rules, and must the document follow the exact template.
+  - DISCUSS: Should the binding document be required to follow W3C copyright rules, and should the document follow the exact template and look and feel.
```
TallTed, line 97:
```suggestion
  - DISCUSS: Should the binding document be required to follow W3C copyright rules, and should the document follow the exact template and look and feel?
```

registry-requirements.md
```diff
@@ -94,9 +94,14 @@ What does the binding have to contain to go into the table
...
     - Ege: No as we want other organizations to also submit bindings. -> Group seems to be fine with this requirement.
-    - Cris: Do we allow the WoT binding to be just one section of the document? -> Group seems to be fine but pointing to the specific location will be needed.
+    - Cris: Do we allow the WoT binding to be just one section of the document? -> Group seems to be fine but pointing to the specific location will be needed. (PDF or similar document types submission will require a text pointing to the section)
+    - Jan: Should the binding document be publically available and for free? What about the license, e.g. can I write a binding driver without any fees etc.?
+      - Ege: At least the custodian and the reviewers should be able to access it for free.
+      - Dimensions: Reading the binding document, reading the protocol specification, implementing a device/Thing, implementing a Consumer application/driver, and building a commercial product with it.
+      - DISCUSS MAIN CALL
+    - Cris: We should strongly recommend linking HTML documents.
```
TallTed, line 104:
```suggestion
    - Cris: Do we allow the WoT binding to be just one section of the document? -> Group seems to be fine but pointing to the specific location will be needed. (PDF or similar document type submission will require a text pointing to the section.)
    - Jan: Should the binding document be publicly available and for free? What about the license, e.g., can I write a binding driver without any fees, etc?
      - Ege: At least the custodian and the reviewers should be able to access it for free.
      - Dimensions: Reading the binding document, reading the protocol specification, implementing a device/Thing, implementing a Consumer application/driver, and building a commercial product with it.
      - DISCUSS MAIN CALL
    - Cris: We should strongly recommend linking HTML documents.
```

TallTed (changes requested):

registry-requirements.md
```diff
@@ -132,3 +132,4 @@ What does the binding have to contain to go into the table
       - Ideally, the custodian, reviewers and submitters would collaborate in a Plugfest on the implementation and collect experience together.
       - Cris: This testing event can continue to collect more inputs on the binding and we showcase these results in a dashboard/report dynamically.
       - DISCUSS: Should we enforce two separate entities?
+        - We can require one Thing and Consumer but collect the amount of implementations and show it in the registry table. More implementations would show more maturity of the binding.
```
TallTed, line 135:
```suggestion
        - We can require one Thing and Consumer but collect the number of independent implementations and show it in the registry table. More implementations would show more maturity of the binding.
```

registry-requirements.md
```diff
@@ -107,22 +108,18 @@ What does the binding have to contain to go into the table
 3. DISCUSS: What is the objective mechanism to confirm the initial entry, i.e., status equals "initial"?
   - Ege: Initial entry is checked for document correctness, i.e., mapping at least one protocol, URI scheme, examples of each mapped operation, "casual" introduction, and abstract. However, we do not check if someone tries to map `readproperty` to a non-existent HTTP method.
   - Cris: Merging the initial entry would trigger a "Call for Implementation". Where discussions on implementation experience should be collected?
-  - Compliance to the point 5 is checked here.
+  - Compliance to the point 6 is checked here.
```
TallTed, line 111:
```suggestion
  - Compliance with point 6 is checked here.
```

registry-requirements.md
```diff
@@ -135,7 +132,16 @@ What does the binding have to contain to go into the table
...
       - DISCUSS: Should we enforce two separate entities?
         - We can require one Thing and Consumer but collect the number of independent implementations and show it in the registry table. More implementations would show more maturity of the binding.
-10. DISCUSS: What are the required sections/content? Is there an order that we require?
-  - Ege: Require an ordering of: Introduction, Binding Content (URL Format, Form Vocabulary Definition as Table, Default and possible mappings to operations as a Table/Template), Examples. The tables should be provided as a template that can be used across different tools (html, word etc.)
+10. The binding MUST contain the following sections in the order presented below. The binding CAN contain other sections anywhere, including between the required ones. The submitters are encouraged to look at the existing submissions. There MUST be at least one operation mapped to a protocol message/vocabulary terms. The submitter SHOULD use the table template provided for the form vocabulary tables in the document.
```
TallTed, line 135:
```suggestion
10. The binding MUST contain the following sections in the order presented below. The binding CAN contain other sections anywhere, including between the required ones. The submitters are encouraged to look at the existing submissions. There MUST be at least one operation mapped to a protocol message/vocabulary term. The submitter SHOULD use the table template provided in the document for the vocabulary.
```

registry-requirements.md
```diff
@@ -135,7 +132,16 @@ What does the binding have to contain to go into the table
...
+  - Introduction
+  - Binding Content:
+    - URL Format
+    - Form Vocabulary Definition as Table,
```
TallTed, line 139:
```suggestion
    - Form Vocabulary Definition as Table
```

registry-requirements.md
```diff
@@ -135,7 +132,16 @@ What does the binding have to contain to go into the table
...
+    - Default and possible mappings to operations as a Table
+  - Examples
+Opinions:
+  - Cris: Fine maybe more sections make sense. To be decided later.
```
TallTed, line 143:
```suggestion
  - Cris: Fine. Maybe more sections make sense. To be decided later.
```

TallTed (changes requested):

registry-requirements.md
```diff
@@ -13,112 +12,130 @@ From the charter:
...
+    - Form Vocabulary Definition as Table
+    - Default and possible mappings to operations as a Table
+  - Examples
+Opinions:
+  - Cris: Fine. Maybe more sections make sense. To be decided later.
+  - Kaz: mention the example bindings as a way how the submitter can write a binding.
+11. DISCUSS: What are the required (machine-readable) documents provided on the side?
+  - Ege: JSON Schema, Ontology File (RDF and HTML) not mandatory?, test report or documentation of implementation experience
+12. DISCUSS: Where should discussions on implementation experience be collected?
+13. DISCUSS: How to handle overlaps between two bindings in the review process?
```
JKRhb, line 141: I was wondering whether the organization that publishes/developed the protocol in question could endorse a certain protocol binding, and that endorsement could then be marked in the table as well (with a checkmark, for example). E.g., OASIS could endorse an/our MQTT binding, making it the “official” one, but there could also be alternative binding documents for MQTT without an endorsement listed in the table.

registry-requirements.md
```diff
@@ -13,112 +12,130 @@ From the charter:
...
+2. A binding that uses a serialization format via the `contentType` keyword MUST mention how the Data Schema terms should be used to describe the messages. This avoids submission of a binding like "XML Binding" that says "Use `contentType:application/xml` and nothing more. That alone would not be enough to serialize correct messages based on the data schema.
+  - TODO: We will need additional mechanisms (including vocabulary terms) to ensure that it is possible to use other media types.
+3. Initial entry MUST be a correct document which complies with point 10 (document content). The reviewer MUST NOT check whether the binding tries to map `readproperty` to a non-existent HTTP method. A successful initial document triggers a "Call for Implementation".
+4. The WoT binding CAN be just one section of the document. In that case, the "Link to the binding document" in the registry entry MUST point to the specific location. PDF or similar document types CAN be submitted if the "Link to the binding document" in the registry entry contains a text pointing to the section. However, HTML and Webpages SHOULD be favoured.
```
JKRhb, line 111: This is a bit nitpicky, but we might want to stick to [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) vocabulary already :) (I.e., replacing CAN with MAY here and in other places).

```suggestion
4. The WoT binding MAY be just one section of the document. In that case, the "Link to the binding document" in the registry entry MUST point to the specific location. PDF or similar document types MAY be submitted if the "Link to the binding document" in the registry entry contains a text pointing to the section. However, HTML and Webpages SHOULD be favoured.
```

registry-requirements.md
```diff
@@ -13,112 +12,130 @@ From the charter:
...
+5. The WoT binding document DOES NOT have to follow the W3C copyright. The submitter is free to choose based on the process they or their organization follows.
+6. The binding document linked in the registry entry DOES NOT have to be open to read, use, implement.
```
JKRhb, line 113: Trying to use the formal keywords here, I was wondering whether we could turn this into a recommendation (to encourage organizations to embrace open standards/specifications) while clarifying that openness is not required.

```suggestion
6. The binding document linked in the registry entry SHOULD be open to read, use, and implement, but it is not required for the document to be added to the registry.
```

registry-requirements.md
egekorkan: I would avoid having two MQTT bindings, one endorsed one not.

JKRhb: Hmm, yeah, you are right of course, I just had the [comment](https://github.com/w3c/wot-binding-templates/issues/393#issuecomment-2622612507) by @benfrancis in https://github.com/w3c/wot-binding-templates/issues/393 in mind here. So essentially, I was wondering whether the authoring organization should also have a say here, or whether the group could also decide to ignore the SDO and recommend a different binding that satisfies the requirements better, for example.

registry-requirements.md
```diff
@@ -13,112 +12,131 @@ From the charter:
...
+  - TODO: We will need additional mechanisms (including vocabulary terms) to ensure that it is possible to use other media types.
+3. Initial entry MUST be a correct document which complies with point 10 (document content). The reviewer MUST NOT check whether the binding tries to map `readproperty` to a non-existent HTTP method. A successful initial document triggers a "Call for Implementation".
+4. The WoT binding MAY be just one section of the document. In that case, the "Link to the binding document" in the registry entry MUST point to the specific location. PDF or similar document types MAY be submitted if the "Link to the binding document" in the registry entry contains a text pointing to the section. However, HTML and Webpages SHOULD be favoured.
+5. The WoT binding document DOES NOT have to follow the W3C copyright. The submitter is free to choose based on the process they or their organization follows.
+6. The binding document linked in the registry entry SHOULD be open to read, use, and implement, but it is not required for the document to be added to the registry.
```
TallTed, line 113:
```suggestion
6. The binding document linked in the registry entry SHOULD be open to read, use, and implement, but it is not required that the document be added to the registry.
```

egekorkan: I think both proposals are changing the meaning. We want to say the openness is not a requirement but is only recommended

JKRhb: Maybe this change could already clarify the bullet point?

```suggestion
6. The binding document linked in the registry entry SHOULD be open to read, use, and implement, but that is not required for the document to be added to the registry.
```

TallTed: I did not understand the intended meaning. Have revised my suggestion (to match that from @JKRhb), now that I *think* I understand the intent.

egekorkan: I have used the suggestion from @TallTed with the revision in the end as that was the initial discussion start. Thanks for the suggestions both!

registry-requirements.md
egekorkan: If the custodian chooses a reviewer that decides on a _better_ one, we have to go with that better one by our rules. So far, we have not declared precedence of an SDO-endorsed over another one. However, given this possibility and comment from @benfrancis, we do have to define what happens in such cases. Created https://github.com/w3c/wot-binding-templates/issues/398 for the discussion. @JKRhb if you are fine with it, let's resolve this comment and continue the discussion there. I am doing that for all discussion points in this pr.

PR #377, opened 2024-07-15
BACnet protocol binding improvements

fennibay:
Hi @egekorkan @ektrah @mjkoster 

Here few improvements for the BACnet protocol binding:

1. The two differences to URI schema in BACnet's Annex Q.8 are listed explicitly -> this is in order to give a concrete idea what the current minimal differences are.
2. Add an example for writing -> was a request by Coleman Brumley back in Feb 2024
3. Change the any model to omit the type field in the WoT data schema. -> it's allowed and it's more correct to cover more complex cases like Variants etc.
4. Auto-format the HTML -> better indentation etc.
5. Use bacv: instead of bacnet: in ttl -> make it consistent with the HTML and JSON schema

I distributed the changes to commits, so should be able to review each change on its own.

PR #379, opened 2024-07-30
BACnet support for events

fennibay:
Bring support for BACnet Events (used interchangeably with Alarms) to WoT BACnet Protocol Binding.

My general approach in this PR is to allow WoT Consumers to interact with EventAffordances in a certain way. I didn't try to cover all features and options of BACnet. I find it better to start simple and add features as they become necessary.

This PR bases on #377, so that one should be merged first.

Preview of rendered doc: <https://deploy-preview-379--wot-binding-templates.netlify.app/bindings/protocols/bacnet/index.html#event-mappings>

### DIFF
--- a/bindings/protocols/bacnet/index.html
+++ b/bindings/protocols/bacnet/index.html
@@ -1018,8 +1018,9 @@ lowercase = %x61-7A  ; "a" to "z"
                         to the same Notification Class. In the typical case, a client will register to all Notification Classes of a BACnet Device.
                     </li>
                     <li>
-                        Some alarms can be configured as acknowledgement required. Acknowledgement is defined by BACnet as a manual action
+                        Some alarms can be configured as "acknowledgement required". Acknowledgement is defined by BACnet as a manual action
                         performed by a human operator, and is separate from the confirmation of the successful reception of the event notification.
+
                     </li>
                 </ol>
                 The picture below demonstrates these key aspects and how they are mapped to the WoT concepts.
--- a/bindings/protocols/bacnet/index.html
+++ b/bindings/protocols/bacnet/index.html
@@ -1126,13 +1126,13 @@ lowercase = %x61-7A  ; "a" to "z"
                                 <li>source: IRI of the BACnet object sending the event</li>
                                 <li>time: Timestamp of the event</li>
                                 <li>prio: Priority of the event</li>
-                                <li>ackReq (optional): Flag indicating if acknowledge is required. Default is false.</li>
+                                <li>ackReq (optional): Flag indicating whether acknowledgement is required. Default is false.</li>
                                 <li>toState: The target state of the event transition</li>
                                 <li>origSource (optional): In case of algorithmic reporting, the IRI of the object monitored by the event-sending object.</li>
-                                <li>eventType (optional): If the event is a regular event transition, an acknowledge notification or a general sync-required event. Default is regular.</li>
+                                <li>eventType (optional): If the event is a regular event transition, an acknowledgement notification or a general sync-required event. Default is regular.</li>
                                 <li>values (optional): Additional values associated with the event, e.g. high-limit in case the value exceeded an OffNormal limit.</li>
                                 <li>message (optional): The message associated with the event.</li>
-                                <li>inactive (optional): Flag if the event is not active anymore, default is false. This is needed for inactive, but not yet acknowledged events; they are there for symmetry with GetCurrentEvents action.</li>
+                                <li>inactive (optional): Flag whether the event is not active anymore. Default is false. This is needed for inactive, but not yet acknowledged events; they are there for symmetry with GetCurrentEvents action.</li>
                             </ul>
                         </td>
                         <td>
@@ -1190,7 +1190,7 @@ lowercase = %x61-7A  ; "a" to "z"
                                         <li>inactive (optional)</li>
                                     </ul>
                                 </li>
-                                <li>moreEvents: Flag indicating if more events should be queried</li>
+                                <li>moreEvents: Flag indicating whether more events should be queried</li>
                             </ul>
                         </td>
                         <td>
--- a/bindings/protocols/bacnet/index.html
+++ b/bindings/protocols/bacnet/index.html
@@ -1029,9 +1029,10 @@ lowercase = %x61-7A  ; "a" to "z"
             <p>
                 The key aspects of the mapping to WoT are:
                 <ol>
-                    <li>Each Notification Class is mapped to an EventAffordance. This maps to the subscription granularity directly.</li>
-                    <li>Event Enrollment objects used for algorithmic reporting are not mapped explicitly, they are per se not relevant to the WoT consumer.</li>
-                    <li>Special event-related actions are mapped to ActionAffordances: getting current events and acknowledging events.</li>
+                    <li>Each Notification Class is mapped to an EventAffordance. This maps directly to the subscription granularity.</li>
+                    <li>Event Enrollment objects used for algorithmic reporting are not explicitly mapped; they are not relevant to the WoT consumer, per se.</li>
+                    <li>Special event-related actions (getting current events and acknowledging events) are mapped to ActionAffordances.</li>
+
                     <li>
                         Thing is assumed to be mapped to the BACnet device. While this is simple, it is not a mandatory decision.
                         In specific cases, one BACnet device can correspond to multiple things, and even vice-versa.
--- a/bindings/protocols/bacnet/index.html
+++ b/bindings/protocols/bacnet/index.html
@@ -1206,9 +1206,9 @@ lowercase = %x61-7A  ; "a" to "z"
                             </ul>
                         </td>
                         <td>
-                            As indicated above, the BACnet payload and the ActionAffordance output differ especially around how events are grouped.
-                            In the BACnet payload, one list item represents one object with its three possible events, which is reflected in Event State, Acknowledged Transitions, Event Timestamps and Priorities.
-                            This is to be processed and split into one standalone event for every current event of this object, or past unacknowledged event, with corresponding timestamp, priority...
+                            As indicated above, the BACnet payload and the ActionAffordance output differ, especially around how events are grouped.
+                            In the BACnet payload, one list item represents one object with its three possible events, which is reflected in Event State, Acknowledged Transitions, Event Timestamps, and Priorities.
+                            This is to be processed and split into one standalone event for every current event, or past unacknowledged event, of this object, with corresponding timestamp, priority, etc.
                         </td>
                     </tr>
                     <tr>
--- a/bindings/protocols/bacnet/index.html
+++ b/bindings/protocols/bacnet/index.html
@@ -1126,13 +1126,13 @@ lowercase = %x61-7A  ; "a" to "z"
                                 <li>source: IRI of the BACnet object sending the event</li>
                                 <li>time: Timestamp of the event</li>
                                 <li>prio: Priority of the event</li>
-                                <li>ackReq (opt.): Flag indicating if acknowledge is required. Default is false.</li>
+                                <li>ackReq (optional): Flag indicating if acknowledge is required. Default is false.</li>
                                 <li>toState: The target state of the event transition</li>
-                                <li>origSource (opt.): In case of algorithmic reporting, the IRI of the object monitored by the event-sending object.</li>
-                                <li>eventType (opt.): If the event is a regular event transition, an acknowledge notification or a general sync-required event. Default is regular.</li>
-                                <li>values (opt.): Additional values associated with the event, e.g. high-limit in case the value exceeded an OffNormal limit.</li>
-                                <li>message (opt.): The message associated with the event.</li>
-                                <li>inactive (opt.): Flag if the event is not active anymore, default is false. This is needed for inactive, but not yet acknowledged events; they are there for symmetry with GetCurrentEvents action.</li>
+                                <li>origSource (optional): In case of algorithmic reporting, the IRI of the object monitored by the event-sending object.</li>
+                                <li>eventType (optional): If the event is a regular event transition, an acknowledge notification or a general sync-required event. Default is regular.</li>
+                                <li>values (optional): Additional values associated with the event, e.g. high-limit in case the value exceeded an OffNormal limit.</li>
+                                <li>message (optional): The message associated with the event.</li>
+                                <li>inactive (optional): Flag if the event is not active anymore, default is false. This is needed for inactive, but not yet acknowledged events; they are there for symmetry with GetCurrentEvents action.</li>
                             </ul>
                         </td>
                         <td>
@@ -1161,7 +1161,7 @@ lowercase = %x61-7A  ; "a" to "z"
                         <td>GetCurrentEvents: ActionAffordance.input</td>
                         <td>
                             <ul>
-                                <li>lastReceivedSource (opt.): IRI of the last received object with event</li>
+                                <li>lastReceivedSource (optional): IRI of the last received object with event</li>
                             </ul>
                         </td>
                         <td>
@@ -1181,13 +1181,13 @@ lowercase = %x61-7A  ; "a" to "z"
                                         <li>source</li>
                                         <li>time</li>
                                         <li>prio</li>
-                                        <li>ackReq (opt.)</li>
+                                        <li>ackReq (optional)</li>
                                         <li>toState</li>
-                                        <li>origSource (opt.)</li>
-                                        <li>eventType (opt.)</li>
-                                        <li>values (opt.)</li>
-                                        <li>message (opt.)</li>
-                                        <li>inactive (opt.)</li>
+                                        <li>origSource (optional)</li>
+                                        <li>eventType (optional)</li>
+                                        <li>values (optional)</li>
+                                        <li>message (optional)</li>
+                                        <li>inactive (optional)</li>
                                     </ul>
                                 </li>
                                 <li>moreEvents: Flag indicating if more events should be queried</li>
--- a/bindings/protocols/bacnet/index.html
+++ b/bindings/protocols/bacnet/index.html
@@ -1013,9 +1013,10 @@ lowercase = %x61-7A  ; "a" to "z"
                         or dedicated other objects can monitor other objects (algorithmic reporting)
                     </li>
                     <li>
-                        Alarms are always sent via Notification Class objects, which keep lists of alarm recipients, i.e. alarm subscriptions
-                        are always done to the Notification Classes, it is not possible to receive alarms only from some of the objects linked
-                        to the same Notification Class. In the typical case, a client will register to all Notification Classes of a BACnet Device.
+                        Alarms are always sent via Notification Class objects, which keep lists of alarm recipients, i.e., alarm subscriptions
+                        are always of the Notification Classes; it is not possible to receive alarms only from some of the objects linked
+                        to a given Notification Class. In the typical case, a client will register to all Notification Classes of a BACnet Device.
+
                     </li>
                     <li>
                         Some alarms can be configured as "acknowledgement required". Acknowledgement is defined by BACnet as a manual action
@@ -1113,7 +1114,8 @@ lowercase = %x61-7A  ; "a" to "z"
                     </tr>
                     <tr>
                         <td>EventAffordance.cancellation</td>
-                        <td>Not modeled, cancellation is implicit</td>
+                        <td>Not modeled; cancellation is implicit</td>
+
                         <td>Same as EventAffordance.subscription</td>
                         <td>Same as EventAffordance.subscription</td>
                     </tr>
--- a/bindings/protocols/bacnet/index.html
+++ b/bindings/protocols/bacnet/index.html
@@ -1664,11 +1664,7 @@ lowercase = %x61-7A  ; "a" to "z"
         </pre>
 
         <p>
-            The resulting example payloads are available below. [[[#example-event-data]]] shows an event notification.
-            [[[#example-acknowledgeevent-input]]] shows what a WoT Consumer would send for acknowledging this event.
-            And [[[#example-getcurrentevents-output]]] shows the same alarm and another alarm that happened before in the
-            GetCurrentEvents action's output, note that in the second alarm <code>ackReq</code> value turned to 0,
-            because now this event was acknowledged.
+            The resulting example payloads are available below.
         </p>
 
         <pre id="example-event-data" class="example" title="Example event data">
@@ -1688,6 +1684,15 @@ lowercase = %x61-7A  ; "a" to "z"
             }
         </pre>
 
+        <p>
+            [[[#example-event-data]]] shows an event notification because of an overheating equipment. Event <code>message</code>
+            gives this as a useful hint, <code>prio</code> is set accordingly. <code>values</code> object gives additional details
+            about the event that the current value is 55 and the exceeded limit was 50. The unit of these values can be found at the
+            <code>source</code> object that triggered this event.
+            Because overheating events are important for the operator to notice, <code>ackReq</code> is set, so the operator definitely
+            will see the event, even after the overheating condition is over.
+        </p>
+
         <pre id="example-acknowledgeevent-input" class="example" title="Example AcknowledgeEvent input">
             {
                 "source": "bacnet://5/0,20",
@@ -1698,6 +1703,11 @@ lowercase = %x61-7A  ; "a" to "z"
             }
         </pre>
 
+        <p>
+            When the operator sees this event and decides to acnowledge it, management station will send the payload shown in
+            [[[#example-acknowledgeevent-input]]]. This can happen before or after the event is over.
+        </p>
+        
         <pre id="example-getcurrentevents-output" class="example" title="Example GetCurrentEvents output">
             {
                 "currentEvents": [
@@ -1726,6 +1736,14 @@ lowercase = %x61-7A  ; "a" to "z"
                 "moreEvents": false
             }
         </pre>
+
+        <p>
+            If the event client wants to get a list of all events of the device, it will invoke the GetCurrentEvents action, and will receive
+            te payload shown in [[[#example-getcurrentevents-output]]]. Here the event notification about the overheating and a past fault event can be
+            seen. This is useful for synchronizing the events known on the client to the events that are on the device, in case of temporary
+            disconnection, reboot of client or device etc. Please note that in the second alarm <code>ackReq</code> value turned to 0,
+            because this event was acknowledged by the operator in the meantime.
+        </p>
 
     </section>
 </body>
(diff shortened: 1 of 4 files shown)


egekorkan:
https://deploy-preview-379--wot-binding-templates.netlify.app/bindings/protocols/bacnet/index.html#event-mappings can be used for seeing the rendered version

ektrah (commented):
Overall this looks good to me, although I haven't checked all the details in the new tables.

However, I think a few more points should be added to make this work within the conceptual framework of WoT. In general, WoT tries to abstract away from the exact protocols used; for a TD consumer, it should not matter whether they use BACnet or another underlying protocol; the drivers take care of the protocol-specific details.

This PR essentially defines a layer above the BACnet Protocol Binding that allows BACnet alarms to be distributed to TD consumers and for them to acknowledge these alarms. While I am not against the BACnet Protocol Binding and this higher layer being specified in the same document, they should be clearly separated and decoupled. Even though the higher layer will usually be used with the BACnet Protocol Binding, it could in principle also be used with the MQTT Protocol Binding, for example.

Let's say a TD consumer sees an event in a TD and it does not look at the contained form(s):

* How does the consumer recognize that this event delivers BACnet events?
    * Should the consumer have the specified data schema hard-coded and compare it to the data schema in the TD, and if they are (almost?) identical, then it's BACnet events?
* How does the consumer recognize that it can/must acknowledge these events?
* How does the consumer know which action to use for confirmation?
    * Should the consumer have the identifier of the action hard-coded?

Let's say I'm implementing a WoT/BACnet driver:

* How does the driver know which BACnet protocol element goes into which key-value pair in the JSON data schema? Is this purely based on the key names?
* I suppose the data schema might evolve with future revisions of the document. What should a driver do if there are newly specified key-value pairs in the data schema that it doesn't yet know about, or if pairs that are currently specified are missing?

Nit:

* You [wrote](https://github.com/w3c/wot-binding-templates/pull/379#issue-2437834798) that the terms "BACnet Event" and "BACnet Alarm" can be used interchangeably. I would suggest using "BACnet Alarm" throughout the document to avoid any confusion with WoT events.

bindings/protocols/bacnet/index.html
```diff
@@ -995,6 +995,244 @@ <h3>WoT Data Schema and BACnet Types</h3>
...
+                            </ul>
+                        </td>
+                        <td>
+                            As indicated above, the BACnet payload and the ActionAffordance output differ especially around how events are grouped.
+                            In the BACnet payload, one list item represents one object with its three possible events, which is reflected in Event State, Acknowledged Transitions, Event Timestamps and Priorities.
+                            This is to be processed and split into one standalone event for every current event of this object, or past unacknowledged event, with corresponding timestamp, priority...
+                        </td>
+                    </tr>
+                    <tr>
+                        <td>AcknowledgeEvent: ActionAffordance.input</td>
```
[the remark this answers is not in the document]
fennibay: I did this now.

bindings/protocols/bacnet/index.html
```diff
@@ -1186,6 +1424,305 @@ <h2 id="examples">Examples</h2>
...
+                            "exceedLim": 50,
+                            "exceedValue": 55,
+                            "statusFlags": [0, 0, 0, 0]
+                        }
+                    }
+                ],
+                "moreEvents": false
+            }
+        </pre>
+
```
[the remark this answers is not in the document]
fennibay: @egekorkan I found a realistic scenario about a breaking fan belt here: <https://sid.siemens.com/r/A6V11602351/22505934987__V6.3_en-US_13226371979>, but that includes non-relevant elements like the additional alarm monitoring logic, operator activating the maintenance switch to replace the fan belt etc. I think it distracts from the actual topic of events.

So, instead I annotated the current overheating example with some text. Please have a look and resolve if OK.

fennibay:
@ektrah thanks for the review and sorry for late reply.

> This PR essentially defines a layer above the BACnet Protocol Binding that allows BACnet alarms to be distributed to TD consumers and for them to acknowledge these alarms. While I am not against the BACnet Protocol Binding and this higher layer being specified in the same document, they should be clearly separated and decoupled. Even though the higher layer will usually be used with the BACnet Protocol Binding, it could in principle also be used with the MQTT Protocol Binding, for example.

Fully agree that the logical payload should be decoupled from the underlying protocol. This is what I tried to express with: "Thus, it is necessary for the WoT consumer to work with a given logical data schema that is influenced by the BACnet standard. Still, this binding takes an approach to define the data schemata of the InteractionAffordances as independent as possible from BACnet, in order to avoid a direct dependency of WoT consumers to a specific protocol."

I tried to handle the decoupling of the logical layer from the protocol binding in the table under the section 6.3, similarly to what we did before in section 6.2. If you have ideas how this can be improved, I'm glad to incorporate them.

> 
> Let's say a TD consumer sees an event in a TD and it does not look at the contained form(s):
> 
> * How does the consumer recognize that this event delivers BACnet events?

I think the goal is that the consumer shouldn't recognize that these are actually BACnet events.

>   
>   * Should the consumer have the specified data schema hard-coded and compare it to the data schema in the TD, and if they are (almost?) identical, then it's BACnet events?

No. The consumer should just work with the schema. If there is an exceptional reason for the consumer to know it's actually BACnet, it should look in the forms.

> * How does the consumer recognize that it can/must acknowledge these events?

Using the ackReq field.

> * How does the consumer know which action to use for confirmation?

There is only one action in the TD: AcknowledgeEvent. This could maybe be improved with a well-known `@type`, but I'm not sure if that's necessary.

>   
>   * Should the consumer have the identifier of the action hard-coded?

Not the id, but the title. Maybe the `@type`

> 
> Let's say I'm implementing a WoT/BACnet driver:
> 
> * How does the driver know which BACnet protocol element goes into which key-value pair in the JSON data schema? Is this purely based on the key names?

Yes, the whole data model is well-known to the BACnet driver, hence it is in this protocol binding spec.

> * I suppose the data schema might evolve with future revisions of the document. What should a driver do if there are newly specified key-value pairs in the data schema that it doesn't yet know about, or if pairs that are currently specified are missing?

This is a general problem for all WoT protocol bindings. If protocol bindings introduce new fields, how is this versioned between the implementations and the binding spec? Is this already conceptualized somewhere?
BACnet binding would follow it.

> 
> Nit:
> 
> * You [wrote](https://github.com/w3c/wot-binding-templates/pull/379#issue-2437834798) that the terms "BACnet Event" and "BACnet Alarm" can be used interchangeably. I would suggest using "BACnet Alarm" throughout the document to avoid any confusion with WoT events.

That is a good idea, I did it now for the places where the term occurs in isolation. However, BACnet has also some phrases in its spec like "event notification", "event state", "event enrollment"; there it would cause more confusion to not use the standard phrase from BACnet, so I kept them.

egekorkan:
I think that @ektrah has many valid points. There are many concepts in this binding that do not exist in the abstraction of WoT. For example, acknowledging an event with an action is not something that we are able to express in a generic way. So saying `Using the ackReq field.` is correct but `ackReq` is not a generic WoT vocabulary for a Consumer to understand. How do we proceed though?

I think this PR should be merged and should be taken as a primary example for manageable affordances discussion (or maybe some state machine modeling of Things). Until we solve that discussion, there will be no way to handle this in a generic fashion. And waiting for that discussion to be resolved is not going to help since that is how BACnet works and not supporting their eventing mechanism would simply mean that BACnet devices with events will not have events in their TD, thus they would be handled out-of-band. This is something WoT standards advocates against, so this is definitely an improvement.

egekorkan:
Call of 06.11:
- It needs to be written in the document that this kind of mechanism require a "BACnet specific Consumer".
- An annotation for the specific actions is preferred so that the Consumer can easily find such important actions related to the event.
- Relationship between the both `"source"` keywords should be explained.
- These explanations should be added as notes in the spec
- People are fine with merging this PR but all agree that a generic mechanism is needed to cover this in the WoT abstraction level, which would work with more protocols.

bindings/protocols/bacnet/index.html
```diff
@@ -1,1192 +1,1751 @@
...
+            <p>
+                With this background, this binding takes some simplifying design decisions:
+                <ol>
+                    <li>
+                        When the BACnet payload is fixed by the BACnet standard for a given service, e.g. for event subscription,
+                        data and alarm acknowledgement, the binding doesn't model the BACnet data with <code>bacnet:hasDataType</code>.
+                    </li>
+                    <li>
+                        Some features of the BACnet alarm model are abstracted away: filtering alarms based on day of week, time of day and event state;
+                        and using unconfirmed event notifications. If these become necessary, they can be added in a future version.
```
TallTed, line 1054:
```suggestion
                        Some features of the BACnet alarm model are abstracted away: filtering alarms based on day of week, time of day, and event state;
                        and using unconfirmed event notifications. If these become necessary, they can be added in a future version.

```

bindings/protocols/bacnet/index.html
```diff
@@ -1,1192 +1,1751 @@
...
+                        <th>Data</th>
+                        <th>WoT Model</th>
+                        <th>BACnet Model</th>
+                        <th>Remarks</th>
+                    </tr>
+                </thead>
+                <tbody>
+                    <tr>
+                        <td>EventAffordance.subscription</td>
+                        <td>Not modeled, subscription is implicit</td>
```
TallTed, line 1080:
```suggestion
                        <td>Not modeled; subscription is implicit</td>

```

bindings/protocols/bacnet/index.html
```diff
@@ -1,1192 +1,1751 @@
...
+                        <td>
+                            The simplifications below allow to omit the WoT Data Model.
+                            <ul>
+                                <li>valid-days &#8592; all days</li>
+                                <li>from-time &#8592; all time</li>
+                                <li>to-time &#8592; all time</li>
+                                <li>recipient &#8592; BACnet address of the WoT Consumer</li>
+                                <li>process-identifier &#8592; BACnet process ID of the WoT consumer</li>
+                                <li>issue-confirmed-notifications &#8592; always true</li>
+                                <li>transitions &#8592; all transitions</li>
```
TallTed, line 1101:
```suggestion
                                <li>valid-days — all days</li>
                                <li>from-time — all time</li>
                                <li>to-time — all time</li>
                                <li>recipient — BACnet address of the WoT Consumer</li>
                                <li>process-identifier — BACnet process ID of the WoT consumer</li>
                                <li>issue-confirmed-notifications — always true</li>
                                <li>transitions — all transitions</li>
```

bindings/protocols/bacnet/index.html
```diff
@@ -1,1192 +1,1751 @@
...
+                            </ul>
+                        </td>
+                    </tr>
+                    <tr>
+                        <td>EventAffordance.cancellation</td>
+                        <td>Not modeled, cancellation is implicit</td>
```
TallTed, line 1107:
```suggestion
                        <td>Not modeled; cancellation is implicit</td>

```

bindings/protocols/bacnet/index.html
```diff
@@ -1,1192 +1,1751 @@
...
+                                <li>time: Timestamp of the event</li>
+                                <li>prio: Priority of the event</li>
+                                <li>ackReq (opt.): Flag indicating if acknowledge is required. Default is false.</li>
+                                <li>toState: The target state of the event transition</li>
+                                <li>origSource (opt.): In case of algorithmic reporting, the IRI of the object monitored by the event-sending object.</li>
+                                <li>eventType (opt.): If the event is a regular event transition, an acknowledge notification or a general sync-required event. Default is regular.</li>
+                                <li>values (opt.): Additional values associated with the event, e.g. high-limit in case the value exceeded an OffNormal limit.</li>
+                                <li>message (opt.): The message associated with the event.</li>
+                                <li>inactive (opt.): Flag if the event is not active anymore, default is false. This is needed for inactive, but not yet acknowledged events; they are there for symmetry with GetCurrentEvents action.</li>
```
TallTed, line 1124:
```suggestion
                                <li>source — IRI of the BACnet object sending the event</li>
                                <li>time — Timestamp of the event</li>
                                <li>prio — Priority of the event</li>
                                <li>ackReq — (optional) Flag indicating whether acknowledgement is required. Default is false.</li>
                                <li>toState: The target state of the event transition</li>
                                <li>origSource — (optional) In algorithmic reporting, the IRI of the object monitored by the event-sending object.</li>
                                <li>eventType — (optional) If the event is a regular event transition, an acknowledgement notification or a general sync-required event. Default is regular.</li>
                                <li>values — (optional) Additional values associated with the event, e.g., high-limit if value exceeded an OffNormal limit.</li>
                                <li>message — (optional) The message associated with the event.</li>
                                <li>inactive — (optional) Flag whether the event is not active anymore. Default is false. This is needed for inactive but not yet acknowledged events; they are there for symmetry with GetCurrentEvents action.</li>
```

bindings/protocols/bacnet/index.html
```diff
@@ -1,1192 +1,1751 @@
...
+                                <li>message-text</li>
+                                <li>notify-type</li>
+                                <li>ack-required</li>
+                                <li>from-state</li>
+                                <li>to-state</li>
+                                <li>event-values</li>
+                            </ul>
+                        </td>
+                        <td>
+                            We focus only on ConfirmedEventNotification, unconfirmed event notifications are currently not supported.
```
TallTed, line 1145:
```suggestion
                            We focus only on ConfirmedEventNotification; unconfirmed event notifications are currently not supported.

```

bindings/protocols/bacnet/index.html
```diff
@@ -1,1192 +1,1751 @@
...
+                        </td>
+                    </tr>
+                    <tr>
+                        <td>GetCurrentEvents: ActionAffordance.input</td>
+                        <td>
+                            <ul>
+                                <li>lastReceivedSource (opt.): IRI of the last received object with event</li>
```
TallTed, line 1152:
```suggestion
                                <li>lastReceivedSource — (optional) IRI of the last received object with event</li>

```

bindings/protocols/bacnet/index.html
```diff
@@ -1,1192 +1,1751 @@
...
+                            </ul>
+                        </td>
+                        <td></td>
+                    </tr>
+                    <tr>
+                        <td>GetCurrentEvents: ActionAffordance.output</td>
+                        <td>
+                            <ul>
+                                <li>
+                                    currentEvents: Array. Each array item is modeled exactly the same as EventAffordance.data:
```
TallTed, line 1167:
```suggestion
                                    currentEvents — Array. Each array item is modeled exactly the same as EventAffordance.data:

```

bindings/protocols/bacnet/index.html
```diff
@@ -1,1192 +1,1751 @@
...
+                                        <li>source</li>
+                                        <li>time</li>
+                                        <li>prio</li>
+                                        <li>ackReq (opt.)</li>
+                                        <li>toState</li>
+                                        <li>origSource (opt.)</li>
+                                        <li>eventType (opt.)</li>
+                                        <li>values (opt.)</li>
+                                        <li>message (opt.)</li>
+                                        <li>inactive (opt.)</li>
```
TallTed, line 1178:
```suggestion
                                        <li>ackReq (optional)</li>
                                        <li>toState</li>
                                        <li>origSource (optional)</li>
                                        <li>eventType (optional)</li>
                                        <li>values (optional)</li>
                                        <li>message (optional)</li>
                                        <li>inactive (optional)</li>
```

bindings/protocols/bacnet/index.html
```diff
@@ -1,1192 +1,1751 @@
...
+                                    </ul>
+                                </li>
+                                <li>moreEvents: Flag indicating if more events should be queried</li>
```
TallTed, line 1181:
```suggestion
                                <li>moreEvents — Flag indicating whether more events should be queried</li>

```

bindings/protocols/bacnet/index.html
```diff
@@ -1,1192 +1,1751 @@
...
+                                        <li>Event Priorities</li>
+                                    </ul>
+                                </li>
+                                <li>More Events</li>
+                            </ul>
+                        </td>
+                        <td>
+                            As indicated above, the BACnet payload and the ActionAffordance output differ especially around how events are grouped.
+                            In the BACnet payload, one list item represents one object with its three possible events, which is reflected in Event State, Acknowledged Transitions, Event Timestamps and Priorities.
+                            This is to be processed and split into one standalone event for every current event of this object, or past unacknowledged event, with corresponding timestamp, priority...
```
TallTed, line 1204:
```suggestion
                            As indicated above, the BACnet payload and the ActionAffordance output differ, especially around how events are grouped.
                            In the BACnet payload, one list item represents one object with its three possible events, which is reflected in Event State, Acknowledged Transitions, Event Timestamps, and Priorities.
                            This is to be processed and split into one standalone event for every current event, or past unacknowledged event, of this object, with corresponding timestamp, priority, etc.
```

bindings/protocols/bacnet/index.html
```diff
@@ -1,1192 +1,1751 @@
...
+                    <tr>
+                        <td>AcknowledgeEvent: ActionAffordance.input</td>
+                        <td>
+                            <ul>
+                                <li>eventSource: Source of the event to be acknowledged</li>
+                                <li>eventState: To-state of the event to be acknowledged</li>
+                                <li>eventTime: Timestamp of the event to be acknowledged</li>
+                                <li>eventTime: Timestamp of the event to be acknowledged</li>
+                                <li>ackSource: String describing the acknowledgement source</li>
+                                <li>ackTime: Timestamp of the acknowledge operation</li>
```
TallTed, line 1216:
```suggestion
                        <td>AcknowledgeEvent — ActionAffordance.input</td>
                        <td>
                            <ul>
                                <li>eventSource — Source of the event to be acknowledged</li>
                                <li>eventState — To-state of the event to be acknowledged</li>
                                <li>eventTime — Timestamp of the event to be acknowledged</li>
                                <li>eventTime — Timestamp of the event to be acknowledged</li>
                                <li>ackSource — String describing the acknowledgement source</li>
                                <li>ackTime — Timestamp of the acknowledge operation</li>
```

bindings/protocols/bacnet/index.html
```diff
@@ -1,1192 +1,1751 @@
...
+                        </td>
+                        <td>
+                            Process Identifier is to be set by WoT Consumer corresponding to what it has entered at subscription time.
+                        </td>
+                    </tr>
+                </tbody>
+            </table>
+            <p>
+                Detailed schemata of these payloads can be found in the JSON Schema linked in this file's header.
+                Examples below shall also be helpful in understanding the details.
```
TallTed, line 1237:
```suggestion
                Examples below will be helpful in understanding the details.

```

bindings/protocols/bacnet/index.html
```diff
@@ -1,1192 +1,1751 @@
...
+            }
+        </pre>
+
+        <p>
+            [[[#example-event-data]]] shows an event notification because of an overheating equipment. Event <code>message</code>
+            gives this as a useful hint, <code>prio</code> is set accordingly. <code>values</code> object gives additional details
+            about the event that the current value is 55 and the exceeded limit was 50. The unit of these values can be found at the
+            <code>source</code> object that triggered this event.
+            Because overheating events are important for the operator to notice, <code>ackReq</code> is set, so the operator definitely
+            will see the event, even after the overheating condition is over.
```
TallTed, line 1693:
```suggestion
            [[[#example-event-data]]] shows an event notification because of an overheating piece of equipment. Event <code>message</code>
            gives this as a useful hint; <code>prio</code> is set accordingly. <code>values</code> object gives additional details
            about the event — that the current value is 55, and the exceeded limit was 50. The unit of these values can be found at the
            <code>source</code> object that triggered this event.
            Because overheating events are important for the operator to notice, <code>ackReq</code> is set, so the operator definitely
            will see the event, even after the overheating condition is over.
```

bindings/protocols/bacnet/index.html
```diff
@@ -1,1192 +1,1751 @@
...
+                "moreEvents": false
+            }
+        </pre>
+
+        <p>
+            If the event client wants to get a list of all events of the device, it will invoke the GetCurrentEvents action, and will receive
+            te payload shown in [[[#example-getcurrentevents-output]]]. Here the event notification about the overheating and a past fault event can be
+            seen. This is useful for synchronizing the events known on the client to the events that are on the device, in case of temporary
+            disconnection, reboot of client or device etc. Please note that in the second alarm <code>ackReq</code> value turned to 0,
+            because this event was acknowledged by the operator in the meantime.
```
TallTed, line 1745:
```suggestion
            If the event client wants to get a list of all events of the device, it will invoke the GetCurrentEvents action, and will receive
            te payload shown in [[[#example-getcurrentevents-output]]]. Here the event notification about the overheating and a past fault event can be
            seen. This is useful for synchronizing the events known on the client to the events that are on the device, in case of temporary
            disconnection, reboot of client or device, etc. Please note that in the second alarm, the <code>ackReq</code> value turned to 0,
            because this event was acknowledged by the operator in the meantime.
```

TallTed (changes requested):

bindings/protocols/bacnet/index.html
```diff
@@ -1,1192 +1,1751 @@
...
+                        </td>
+                    </tr>
+                </tbody>
+            </table>
+        </section>
+        <section id="event-mappings">
+            <h3>Mappings related to events</h3>
+            <p>
+                BACnet has an extensive alarm model, see [[BACnet]] Chapter 13. Alarm and Event Services for a complete description.
+                The fundamental aspects are:
```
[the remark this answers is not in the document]
fennibay:
```suggestion
                The fundamental aspects are as follows:

```

bindings/protocols/bacnet/index.html
```diff
@@ -1,1192 +1,1751 @@
...
+                        and Fault (e.g. defect sensor). Events are sent in case of transitions, including self-transitions and toNormal transitions.
+                    </li>
+                    <li>
+                        Objects can monitor themselves for event states (intrinsic reporting),
+                        or dedicated other objects can monitor other objects (algorithmic reporting)
+                    </li>
+                    <li>
+                        Alarms are always sent via Notification Class objects, which keep lists of alarm recipients, i.e. alarm subscriptions
+                        are always done to the Notification Classes, it is not possible to receive alarms only from some of the objects linked
+                        to the same Notification Class. In the typical case, a client will register to all Notification Classes of a BACnet Device.
```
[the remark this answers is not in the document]
fennibay: You used subscribe "of" in the second line, instead I'd like to keep using "to". Something like this:

```suggestion
                        Alarms are always sent via Notification Class objects, which keep lists of alarm recipients, i.e. alarm subscriptions
                        are always to the Notification Classes, it is not possible to receive alarms only from some of the objects linked
                        to the same Notification Class. In the typical case, a client will subscribe to all Notification Classes of a BACnet Device.

```

bindings/protocols/bacnet/index.html
fennibay: I use the colon character to indicate I am defining the meaning of these terms, like a vocabulary notation. Is there a specific reason to use the dash instead?

fennibay:
Discussion with @egekorkan and @ektrah today:

 
> * It needs to be written in the document that this kind of mechanism require a "BACnet specific Consumer".

We will formulate this differently, because the payload is intended to be abstracted from BACnet, so also a WoT Consumer without knowledge of BACnet can work with events.

However, some implicit knowledge is required to link `ackReq` to `AcknowledgeEvent`, which is currently specified only in the BACnet protocol bindings. This will be documented in the EventAffordance short-term, and the long-term solution is "manageable affordances", see below.

> * An annotation for the specific actions is preferred so that the Consumer can easily find such important actions related to the event.
> * Relationship between the both `"source"` keywords should be explained.
> * These explanations should be added as notes in the spec

I will add these both for `GetCurrentEvents` and `AcknowledgeEvent`, and provide a link to [manageable affordances](https://github.com/w3c/wot-thing-description/blob/main/planning/work-items/new-features.md#manageable-affordances)

Example for note: <https://github.com/w3c/wot-thing-description/blob/main/index.html#L1704>

bindings/protocols/bacnet/index.html
TallTed: It's a stronger visual separator. Also, a number of tools (wrongly) include the colon when double-clicking to select the term, but they don't include the em-dash (because it's separated by a space). (I believe this select behavior is part of why programmers often insert a space before a colon, which space is itself incorrect.)

bindings/protocols/bacnet/index.html
fennibay: I kept the colon, took over the wording.

bindings/protocols/bacnet/index.html
fennibay: Change done.

bindings/protocols/bacnet/index.html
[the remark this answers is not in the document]
fennibay: @egekorkan assuming this one is OK and resolving.

fennibay:
@egekorkan @ektrah I added the note now, as discussed yesterday, please have a look.

bindings/protocols/bacnet/index.html
```diff
@@ -1032,8 +1032,27 @@ <h3>Mappings related to events</h3>
...
                     <li>Each Notification Class is mapped to an EventAffordance. This maps directly to the subscription granularity.</li>
                     <li>Event Enrollment objects used for algorithmic reporting are not explicitly mapped; they are not relevant to the WoT consumer, per se.</li>
-                    <li>Special event-related actions (getting current events and acknowledging events) are mapped to ActionAffordances.</li>
-
+                    <li>
+                        Special event-related actions (getting current events and acknowledging events) are mapped to ActionAffordances.
+                        <div class="note">
+                            <p>The WoT consumer does not need to know the details of BACnet alarming, it can simply interact with the InteractionAffordances of the Thing,
+                            as per their data schemata. However the consumer still needs to have an implicit understanding about the relations
+                            between the InteractionAffordances:</p>
```
TallTed, line 1040:
```suggestion
                            <p>The WoT consumer does not need to know the details of BACnet alarming; it can simply interact with the InteractionAffordances of the Thing,
                            according to their data schemata. However, the consumer does need to have an understanding of the relations
                            between the InteractionAffordances:</p>

```

TallTed (changes requested):

bindings/protocols/bacnet/index.html
```diff
@@ -1032,8 +1032,27 @@ <h3>Mappings related to events</h3>
...
+                            <ol>
+                                <li>
+                                    <code>GetCurrentEvents</code> action can be used to fetch all current events of the thing, which would not be received
+                                    just by subscribing to events. This action delivers the same payload as the EventAffordance's data.
```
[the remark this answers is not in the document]
fennibay: I used the code tag for terms occurring in JSON or other code. So here I leave it out.

egekorkan:
As reviewed in the call, the review points are incorporated and the PR can be merged

PR #380, opened 2024-10-08
HTTP Binding to TD Alignment

egekorkan:
As mentioned in https://github.com/w3c/wot-thing-description/issues/1259#issuecomment-2289023551, I am starting the work to move the missing content here. I am copying the comment content below and marking them if they are already addressed in the PR. Some points require discussion and this comment will be updated as the PR is updated.

- [ ] 1. TD mentions the default inclusion of the HTTP vocabulary in the TD Context. We need to discuss if we should continue doing this.
- [x] 2. TD mentions the default mechanism with an assertion `When no method is indicated in a form representing an Protocol Binding based on HTTP, a Default Value MUST be assumed as shown in the following table.` This should be moved to HTTP Binding document. I have thought of having a generic assertion about this but we would test each binding on its own and should have a corresponding assertion in each document. -> There is no assertion but the text is moved
- [x] 3. TD default values table have methods in the first column with multiple operations on the right cell. The way it is done in the binding makes more sense to me, i.e. mapping an operation to the method.
- [x] 4. TD has [an example](https://w3c.github.io/wot-thing-description/#example-58) that shows default values insertion
- [x] 5. TD has an assertion (but not RFC!) saying `In the case of a forms entry that has multiple op values the usage of the htv:methodName is not permitted. ` This should be turned to a real assertion in binding and also should be mentioned in a generic way in the binding registry as it applies to all protocols. This also has a corresponding interactive example. -> There is no assertion but the text is moved
- [x] 6. Binding has `fieldName` and `fieldValue` in the vocabulary term table. This should be kept.
- [ ] 7. Binding has sequence diagram examples. I see some issues with WebSocket examples since an unsubscription is not necessarily disconnecting the connection. Otherwise they are fine.
- [x] 8. Binding refers to TD1.0 REC instead of 1.1

Otherwise, I have done the following:

- Add checkbox code
- Add conformance section
- Add terminology section

### DIFF
--- a/bindings/protocols/http/index.html
+++ b/bindings/protocols/http/index.html
@@ -55,6 +55,88 @@
             ]
         };
     </script>
+    <style>
+        a[href].internalDFN {
+            color: inherit;
+            border-bottom: 1px solid #99c;
+            text-decoration: none;
+        }
+    
+        .example {
+            border-color: #ea1252;
+            background: #fef11e;
+            counter-increment: example;
+            overflow: auto;
+            clear: both;
+        }
+    
+        .with-default .with-default-control {
+            margin-top: 1em;
+            margin-bottom: 1em;
+        }
+    
+        .with-default pre.selected {
+            margin-top: 0;
+            margin-bottom: 1em;
+            height: auto;
+        }
+    
+        .with-default pre {
+            margin-top: 0;
+            margin-bottom: 0;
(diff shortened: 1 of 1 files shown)


egekorkan:
I will keep it open for a couple of days, but there were no concerns raised in the call of 10th of October.

PR #381, opened 2024-10-29
fix: add day of week to BACnet protocol binding's Date and DateTime types

fennibay:
Without wildcards, day of week would not be necessary, that's why ISO8601 doesn't support day of week in [calendar dates](https://en.wikipedia.org/wiki/ISO_8601#Calendar_dates).

Because BACnet supports wildcards for the dates, we need to represent day of week. This change takes the day of week representation of ISO8601's [week dates](https://en.wikipedia.org/wiki/ISO_8601#Week_dates) and extends the ISO8601's calendar date format with it.

### DIFF
--- a/bindings/protocols/bacnet/index.html
+++ b/bindings/protocols/bacnet/index.html
@@ -656,9 +656,10 @@ lowercase = %x61-7A  ; "a" to "z"
                                 {
                                     "type": "string",
                                     "pattern": "^((([1-9][0-9]*)?[0-9]{4})|\\*)-((1[0-2]|0[1-9])|\\*)-((3[01]|0[1-9]|[12][0-9])|\\*)(\\s*\\([1-7\\*]\\))?T((2[0-3]|[01][0-9])|\\*):([0-5][0-9]|\\*):([0-5][0-9]|\\*)(\\.[0-9]+)?(Z|[+-]\\d\\d:\\d\\d)?$",
-                                    "$comment": "ISO8601 Date Time Format, extended with day of week (1: Mon, 7: Sun) and with optional wildcards"
+                                    "$comment": "ISO8601 Date Time Format, optionally extended with day of week (1: Mon, 7: Sun) and with optional wildcards"
                                 }
                             </code>
+                            Please see <code>bacv:Date</code> and <code>bacv:Time</code> about the use of wildcards.
                         </td>
                         <td>
                             <code>
@@ -746,9 +747,15 @@ lowercase = %x61-7A  ; "a" to "z"
                                 {
                                     "type": "string",
                                     "pattern": "^((([1-9][0-9]*)?[0-9]{4})|\\*)-((1[0-2]|0[1-9])|\\*)-((3[01]|0[1-9]|[12][0-9])|\\*)(\\s*\\([1-7\\*]\\))?$",
-                                    "$comment": "ISO8601 Date Format, extended with day of week (1: Mon, 7: Sun) and with optional wildcards"
+                                    "$comment": "ISO8601 Date Format, optionally extended with day of week (1: Mon, 7: Sun) and with optional wildcards"
                                 }
                             </code>
+
+                            The wildcards are used to indicate simple recurrences, e.g.
+                            <code>*-*-1</code> would indicate the first day of each month, and <code>*-6-* (1)</code> would indicate all Mondays in June.
+                            It is possible to indicate an empty set by giving an invalid day of week such as <code>2024-11-07 (1)</code>
+                            (actually it is a Thursday), the more straightforward way would be to just leave out this entry where this data would occur.
+                            All values may be set to wildcards <code>*-*-* (*)</code> to indicate any date.
                         </td>
                         <td>
                             <code>
@@ -770,6 +777,10 @@ lowercase = %x61-7A  ; "a" to "z"
                                     "$comment": "ISO8601 Time Format with optional wildcards"
                                 }
                             </code>
+
+                            The wildcards are used to indicate occurrences,
+                            e.g. <code>*:00:00Z</code> to indicate all seconds of the first minute of the hour in UTC.
+                            <code>*:*:*</code> can be used to indicate any time.
                         </td>
                         <td>
                             <code>
--- a/bindings/protocols/bacnet/index.html
+++ b/bindings/protocols/bacnet/index.html
@@ -655,11 +655,12 @@ lowercase = %x61-7A  ; "a" to "z"
                             <code>
                                 {
                                     "type": "string",
-                                    "pattern": "^((([1-9][0-9]*)?[0-9]{4})|\\*)-((1[0-2]|0[1-9])|\\*)-((3[01]|0[1-9]|[12][0-9])|\\*)(\\s*\\([1-7\\*]\\))?T((2[0-3]|[01][0-9])|\\*):([0-5][0-9]|\\*):([0-5][0-9]|\\*)(\\.[0-9]+)?(Z|[+-]\\d\\d:\\d\\d)?$",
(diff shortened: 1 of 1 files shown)


fennibay:
@egekorkan @ektrah @mjkoster here a small change, please review.

ektrah (approved):
I've tried to come up with a few examples and how I'd interpret them:

- `"*-*-01"` -- the set of dates containing the first day of every month
- `"2024-*-01"` -- the set of dates containing the first day of every month in 2024
- `"*-*-*(4)"` -- the set of dates containing every Thursday of every month
- `"*-10-*(4)"` -- the set of dates containing every Thursday in October every year
- `"*-10-29(4)"` -- the set of dates containing the 29th day of October every year, if it is a Thursday
- `"2024-10-29(4)"` -- the single-element set consisting of the 29th day of October in the year 2024
- `"2024-10-29(7)"` -- the empty set

Is that the correct interpretation?

Since the wildcards are not part of ISO8601, should they maybe be explained in a bit more detail? (perhaps in a separate pull request)

Generally speaking, would it make sense to re-use something like iCalendar recurrence rules or crontab syntax to express recurrence like these, instead of inventing a new syntax?

Otherwise this looks good to me.

fennibay:
Thanks for the comments @ektrah 

Your examples are correct. I can explain the wildcards a bit more, ultimately BACnet standard is the main source here, the mappings only reflect what BACnet can represent in date/time formats. Basically an asterisk means it can be any year/month/day/day-of-week, and that's it. Does that need further explanation?

I checked crontab and iCalendar in detail:
- [crontab](https://en.wikipedia.org/wiki/Cron) doesn't support specific year, so that would be an extension, as we are now doing for ISO-8601 with asterisks
- iCalendar, [RFC2445](https://www.[rfc-editor.org/rfc/rfc2445.txt](https://www.rfc-editor.org/rfc/rfc2445.txt): Here I was a bit lost in the spec. This is much more capable than BACnet Date/Time representations, so it is much more complicated. It is also using some English keywords like WEEKLY, which would be a language-specific representation. And most critically we would have to define complex mappings going from 2024-*-* to something like all days in 2024, so DAILY with start and end dates, with the start and stop dates of 2024-01-01 and 2024-12-31. Or do you see a simpler way to map?

egekorkan (approved):

egekorkan:
If the discussion above is finalized and the changes are ok with @mjkoster , it can be merged but we will check in today's call as all the approvals are from the same organization

ektrah:
I think the important insight here is that we are not describing a date (or date and time), but a _recurrence rule_. In other words, the string specifies a set of dates (or dates and times) &ndash; even if no wildcards are present. And counterintuitively, strings with an invalid date (or date and time) are not actually invalid, since they simply describe the empty set. I think this would be very helpful to point out in the document.

Too bad that the crontab syntax does not support years, otherwise it would have been nice to reuse parts of the syntax instead of defining a new one. I agree with you that iCalendar is too complex for our purposes.

fennibay:
@ektrah ok your point makes sense. I will explain this generally for the date / time types and add some examples. Then I hope we can merge.

egekorkan:
Reviewed in the call: We are fine with the changes. We can merge it once the textual improvements are done.

fennibay:
@egekorkan @ektrah I did the textual changes now. I realized I hadn't added the odd and even wildcards for month and day of month, so now they are also in. Please have a last look before merging.

bindings/protocols/bacnet/index.html
```diff
@@ -770,6 +792,10 @@ <h3>WoT Data Schema and BACnet Types</h3>
                                     "$comment": "ISO8601 Time Format with optional wildcards"
                                 }
                             </code>
+
+                            The wildcards are used to indicate occurrences,
+                            e.g. <code>*:00:00Z</code> to indicate all seconds of the first minute of the hour in UTC.
```
TallTed, line 797: This doesn't make sense to me. I see "the zero'th second of every hour". My suggestion *should* indicate every second of the first minute of each hour.
(Also, there's a missing comma.)
```suggestion
                            Wildcards are used to indicate occurrences,
                            e.g., <code>*:00:*Z</code> to indicate all seconds of the first minute of each hour in UTC.
```

bindings/protocols/bacnet/index.html
```diff
@@ -745,10 +747,30 @@ <h3>WoT Data Schema and BACnet Types</h3>
...
+                                <li><code>O</code> and <code>E</code> can be used in the day of week field to indicate odd and even days of the month.</li>
+                                <li><code>*</code> can be used for any year, month, day of month or day of week.</li>
+                                <li>Examples
+                                    <ul>
+                                        <li><code>*-*-1</code>, first day of each month</li>
+                                        <li><code>*-6-* (1)</code>, all Mondays in June.</li>
+                                        <li><code>*-E-1</code>, first day of all even months.</li>
+                                        <li>
+                                            <code>2024-11-07 (1)</code>, an empty set, because day of week is actually Thursday (4),
+                                            the more straightforward way would be to just leave out this entry where this data would occur.
```
TallTed, line 766:
```suggestion
                                            <code>2024-11-07 (1)</code>, an empty set, because the day of week for this date is actually Thursday (4).
                                            A more straightforward way would be to just leave out the entry where this data would occur.
```

bindings/protocols/bacnet/index.html
```diff
@@ -745,10 +747,30 @@ <h3>WoT Data Schema and BACnet Types</h3>
...
+                                        </li>
+                                        <li><code>2024-11-07 (4)</code> / <code>2024-11-07</code>, an exact date with or without the day of week.</li>
+                                        <li><code>2024-11-07 (4)</code> / <code>2024-11-07</code>, an exact date with or without the day of week.</li>
+                                        <li><code>*-*-* (*)</code> / <code>*-*-*</code>, any date</li>
```
ektrah, line 770:
```suggestion
                                        <li><code>*-*-1</code>: Recurs on the 1st day of every month, regardless of year.</li>
                                        <li><code>*-6-* (1)</code>: Recurs on every Monday in June, regardless of year.</li>
                                        <li><code>*-E-1</code>: Recurs on the 1st day of every even month, regardless of year.</li>
                                        <li>
                                            <code>2024-11-07 (1)</code>: November 7, 2024, is not a Monday, resulting in no recurrences.
                                        </li>
                                        <li><code>2024-11-07 (4)</code> / <code>2024-11-07</code>: November 7, 2024, is a Thursday, resulting in a single recurrence.</li>
                                        <li><code>*-*-* (*)</code> / <code>*-*-*</code>: Recurs every day, regardless of year, month, or day of the week.</li>
```

ektrah (changes requested):
- `|`-separators are missing in the regular expressions.
- Some examples do not match the regular expressions.
- A few suggestions for the text.

bindings/protocols/bacnet/index.html
```diff
@@ -770,6 +792,10 @@ <h3>WoT Data Schema and BACnet Types</h3>
                                     "$comment": "ISO8601 Time Format with optional wildcards"
                                 }
                             </code>
+
+                            The wildcards are used to indicate occurrences,
+                            e.g. <code>*:00:00Z</code> to indicate all seconds of the first minute of the hour in UTC.
+                            <code>*:*:*</code> can be used to indicate any time.
```
ektrah, line 798:
```suggestion
                            Examples:
                            <ul>
                                <li><code>*:00:00Z</code>: Recurs every hour on the hour (00 minutes, 00 seconds), using Coordinated Universal Time (UTC).</li>
                                <li><code>*:*:*</code>: Recurs every second of every minute and every hour, essentially occurring continuously throughout the day.</li>
                                <li><code>06:00:00Z</code>: Occurs at 6:00 AM UTC, with 0 minutes and 0 seconds past the hour.</li>
                            </ul>
```

bindings/protocols/bacnet/index.html
fennibay: Thank you, this was a mistake. Now corrected with the suggestion from @ektrah.

fennibay:
@egekorkan I think I was able to incorporate all findings. Please merge when you get a chance.

@ektrah @TallTed many thanks for the feedback.

ektrah (approved):

Issue #382, opened 2024-11-20
Syntax error in Bacnet ontology


mahdanoura:
https://github.com/w3c/wot-binding-templates/blob/07230b73ccefaf99d5bda8ede6a1c354989800be/bindings/protocols/bacnet/bacnet-model.ttl#L192

egekorkan:
@fennibay FYI :)

fennibay:
@mahdanoura thanks for catching!

The fix is in #383, @egekorkan please merge. Or should I just push such small fixes directly to main?

fennibay:
#383 merged, closing.

PR #383, opened 2024-11-21
fix: syntax error in BACnet ontology

fennibay:
#382


ektrah (approved):

egekorkan (approved):
I think this is all that is expected to be fixed right? The PR is in draft status but I guess it is a mistake. In any case, such a small fix can merged async

Issue #386, opened 2024-12-02
[BACnet] Type and range of bacv:usesService


danielpeintner:
The term [`bacv:usesService`](https://w3c.github.io/wot-binding-templates/bindings/protocols/bacnet/index.html#vocabulary-form-terms) is defined as an enum of "ReadProperty", "WriteProperty", "SubscribeCOV" with type string.

Note1: I think with events there is also "AcknowledgeAlarm" and "GetEventInfo" ?

Note2: (some) examples use wrong prefix
e.g., `"bacnet:usesService"` instead of `"bacv:usesService"`

Note3: According to [example 6](https://w3c.github.io/wot-binding-templates/bindings/protocols/bacnet/index.html#example-event) it can also be of type array. Not only string.

ektrah:
> Note1: I think with events there is also "AcknowledgeAlarm" and "GetEventInfo" ?

That's right, these should be added.

> Note2: (some) examples use wrong prefix e.g., `"bacnet:usesService"` instead of `"bacv:usesService"`

This needs to be fixed.

> Note3: According to [example 6](https://w3c.github.io/wot-binding-templates/bindings/protocols/bacnet/index.html#example-event) it can also be of type array. Not only string.

As I understand it, JSON-LD in general does not distinguish between a string and a single element array containing a string. So I guess this is where this comes from. However, you are right that in a TD we should be a bit more explicit/restrictive. I think it is not necessary to be able to specify multiple services, so the examples should be fixed to use a string instead of an array of strings.

danielpeintner:
> > Note3: According to [example 6](https://w3c.github.io/wot-binding-templates/bindings/protocols/bacnet/index.html#example-event) it can also be of type array. Not only string.
> 
> As I understand it, JSON-LD in general does not distinguish between a string and a single element array containing a string. So I guess this is where this comes from. However, you are right that in a TD we should be a bit more explicit/restrictive. I think it is not necessary to be able to specify multiple services, so the examples should be fixed to use a string instead of an array of strings.

I think we once had the idea/plan that each binding also provides a JSON schema with the extensions.
A JSON schema would definitely need to know whether it is a string, an array, or both.

@egekorkan did we drop that "requirement" ?

ektrah:
As said above, we can and should be more restrictive in TDs than general JSON-LD. I was just trying to explain where the array in the examples might come from. Instead of amending the JSON schema to allow arrays in `bacv:usesService`, the examples should be fixed.

[...]

PR #388, opened 2024-12-03
fix: bacv:usesService fixes

fennibay:
Update the range in the documentation.
In the examples:
- Correct the relicts of bacnet: prefix to bacv:
- Fix to always have a single element in the value, not an array

Type and range of bacv:usesService #386

/cc @ektrah @danielpeintner


ektrah (approved):

egekorkan (approved):

danielpeintner (approved):
LGTM

Issue #357, opened 2024-04-01
BACnet media type


egekorkan:
In the BACnet binding, there is no usage of `contentType` in the examples so the default mechanism would apply and we would get JSON payloads. However, BACnet is not using JSON afaik. How should we handle this @fennibay ?

fennibay:
I understand that the logical payload is always JSON-based. contentType defines the physical payload on the wire.

BACnet always uses the same ASN.1-based encoding on the wire. It is implemented in the available BACnet stacks.

So from BACnet POV there is not really a need for a contentType field.

But for the sake of completeness we could define one as a default in schema. I checked the [media types on IANA](https://www.iana.org/assignments/media-types/media-types.xhtml), and I don't see anything for BACnet or ASN.1. So we would have to register one on IANA first.

There is [application/bacnet-xdd+zip](https://www.iana.org/assignments/media-types/application/bacnet-xdd+zip), but that one is not for BACnet data, but a special BACnet file format, used for exchanging device descriptions.

egekorkan:
> So from BACnet POV there is not really a need for a contentType field.

The issue is that if we don't define one, a TD process MUST automatically insert application/json (see [here](https://w3c.github.io/wot-thing-description/#sec-default-values)).

The contentType in a TD form doesn't have to be registered in IANA first (at least for now) so it should be fine to define one in the binding spec for now.

fennibay:
Ok, so can we define it like this?

"contentType": "application/X-bacnet"

danielpeintner:
Minor: I usually see lower-case content-types only...

Even though looking int https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.1 states some vague statements like

> The type, subtype, and parameter name tokens are case-insensitive.
> Parameter values might or might not be case-sensitive, depending on
> the semantics of the parameter name.

egekorkan:
I propose to use `"application/bacnet-x"` to be aligned with lowercase throughout and starting with bacnet since the other bacnet media type starts like that too. Ideally, something should be registered to IANA. Any thoughts @ektrah and @mjkoster ?

ektrah:
[...]

PR #390, opened 2024-12-16
fix: specify BACnet contentType

fennibay:
Without specification it falls back to application/json, which is wrong
Now it will default to application/octet-stream, which means arbitrary binary data

BACnet media type #357

### DIFF
--- a/bindings/protocols/bacnet/bacnet.schema.json
+++ b/bindings/protocols/bacnet/bacnet.schema.json
@@ -24,6 +24,10 @@
                 },
                 "bacv:hasDataType": {
                     "$ref": "#/definitions/bacnetDataType"
+                },
+                "contentType": {
+                    "const": "application/octet-stream",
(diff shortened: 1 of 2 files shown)


bindings/protocols/bacnet/bacnet.schema.json
```diff
@@ -24,6 +24,10 @@
                 },
                 "bacv:hasDataType": {
                     "$ref": "#/definitions/bacnetDataType"
+                },
+                "contentType": {
+                    "const": "application/octet-stream",
+                    "default": "application/octet-stream"
```
egekorkan, line 30:
```suggestion
                "contentType": {
                    "const": "application/octet-stream"
```

egekorkan (approved):
I was going to remove the default from the `contentType` but it actually makes sense if a bacnet TD is processed, where a JSON Schema validator can insert it if needed.

Issue #389, opened 2024-12-07
CoAP: Vocabulary for (non-)confirmable messages and retry patterns


JKRhb:
During WoT Week, @mkovatsc brought up that the CoAP Binding Template currently does not provide a way to specify that requests sent to a Thing should be non-confirmable, e.g., because it is a sleepy device (see https://github.com/w3c/wot-thing-description/issues/2059) that is only available within a certain time frame. This could be covered by a new boolean term (e.g., `cov:confirmable`), where the absence of the term would imply that the consumer is free to choose the kind of request it wants to send. 

Related to that, it might be useful to be able to specify the pattern consumers should use when trying to reach the device in question, e.g., whether they should rather retry sending a non-confirmable request with a certain interval. This could be useful to have if the consumer is not able to determine the exact point in time when the Thing will be reachable again. Not only because I am not entirely sure how to model this vocabulary term exactly, this will probably require some discussion, though.

danielpeintner:
While I think we have a use-case for CoAP, I am pretty sure other devices/protocols (that are designed to last for a long time, like LoRaWAN) would benefit from a similar term.
Hence, it makes me think whether we could introduce a more generic term to describe this aspect...

PR #392, opened 2024-12-19
feat: introduce cov:confirmable vocabulary term

JKRhb:
This PR contains my first attempt at resolving one of the aspects mentioned in #389 by introducing a `cov:confirmable` vocabulary term.


ektrah:
I think it would be helpful to explain _when_ to include a `"cov:confirmable": true` or `"cov:confirmable": false` in a TD (e.g., by giving some  examples and/or a rule of thumb). Otherwise, this looks good to me.

ektrah (changes requested):
I think it would be helpful to explain _when_ to include a `"cov:confirmable": true` or `"cov:confirmable": false` in a TD (e.g., by giving some  examples and/or a rule of thumb). Otherwise, this looks good to me.

JKRhb:
> I think it would be helpful to explain _when_ to include a `"cov:confirmable": true` or `"cov:confirmable": false` in a TD (e.g., by giving some examples and/or a rule of thumb). Otherwise, this looks good to me.

Thank you for your feedback, @ektrah! I tried to give some examples in 3df913868aeae8890f38c31d16b419cf99b47702, let me know if you see room for improvement here :)

bindings/protocols/coap/index.html
```diff
@@ -210,6 +210,56 @@ <h3>Requests/Responses</h3>
     }
 }
 </pre>
+            <p>
+                In CoAP over UDP or DTLS [[RFC7252]], requests and responses can either be confirmable (CON) or non-confirmable (NON).
+                In contrast to non-confirmable messages, confirmable messages are required to be acknowledged by the receiving endpoint and are otherwise retransmitted until an acknowledgement has been received.
+                Using non-confirmable requests can be useful for sensor data (that gets sent repeatedly and therefore does not need to be transmitted reliably), while confirmable requests provide a certain reliability that is useful for retrieving values or invoking actions that should not fail.
```
ektrah, line 216: In a TD we want to tell the TD consumer how to interact with the Thing, so this is about CoAP requests that the consumer sends to the Thing, right?

I would assume that if a TD consumer knows what it is doing (e.g. transmitting sensor data, retrieving a value, or invoking an action) it also knows whether it is best to do it with a CON or NON. Maybe the consumer just needs to give the CoAP driver a hint as to what it is doing?

Or what are the cases where a TD consumer is doing something and also knows what it is doing, but does not know whether to use CON or NON, and so the TD should tell it (or overrule what it would normally do)?

JKRhb: Hmm, yeah, these are good points. Maybe a better example could be sleepy devices that were also discussed in the context of #389? Although I am still not completely sure how to model that. Perhaps another use case could be resource constraints in general, where using NON could be slightly more efficient (although not that much when piggybacked acknowledgments are being used).

Do you see a better motivation here? (Also pinging @mkovatsc as we discussed this potential addition to the vocabulary during WoT week.)

ektrah: Sleepy devices are interesting because a TD consumer may spend a lot of time trying to reach a device that is unreachable, draining the resources of the network infrastructure, which may also be resource constrained. It would therefore be very desirable if a TD could give a consumer hints on how best to reach the sleepy device. Is the choice between CON and NON enough? Is the sleepy device awake at regular intervals so that the TD consumer only has to synchronize its schedule once and can then communicate at the fixed interval that is specified in the TD? This would then no longer be CoAP-specific but would go in the direction of the generic terms mentioned by @danielpeintner.

bindings/protocols/coap/index.html
```diff
@@ -210,6 +210,56 @@ <h3>Requests/Responses</h3>
...
+            </p>
+            <p>
+                To provide a hint whether CON or NON requests should be used for accessing a CoAP resource associated with a form, the vocabulary term `cov:confirmable` can be used.
+                A value of `true` indicates that a CON request SHOULD be used by the Consumer, while a value of `false` indicates that a NON request SHOULD be used instead (for potential usages of the vocabulary term see [[[#example-request-con-non]]]).
+                However, as the vocabulary term is just a hint, the consumer is not bound to this recommendation.
+                When `cov:confirmable` is absent from a CoAP form, the consumer can choose freely between CON and NON requests as there is no default value defined.
```
TallTed, line 222: A "hint" seems rather less powerful than the SHOULD (which looks like RFC2119, to me).
```suggestion
                The vocabulary term `cov:confirmable` can be used to provide a hint whether CON or NON requests are expected to be better for accessing a CoAP resource associated with a form.
                A value of `true` encourages the Consumer to use CON requests, while a value of `false` encourages the Consumer to use NON requests. (For potential uses of the vocabulary term, see [[[#example-request-con-non]]].)
                Note that, as the vocabulary term is just a hint, the consumer is not bound to this recommendation.
                When `cov:confirmable` is absent from a CoAP form, the consumer can choose freely between CON and NON requests as there is no default value defined.
```

bindings/protocols/coap/index.template.html
```diff
@@ -210,6 +210,56 @@ <h3>Requests/Responses</h3>
...
+                In contrast to non-confirmable messages, confirmable messages are required to be acknowledged by the receiving endpoint and are otherwise retransmitted until an acknowledgement has been received.
+                Using non-confirmable requests can be useful for sensor data (that gets sent repeatedly and therefore does not need to be transmitted reliably), while confirmable requests provide a certain reliability that is useful for retrieving values or invoking actions that should not fail.
+            </p>
+            <p>
+                To provide a hint whether CON or NON requests should be used for accessing a CoAP resource associated with a form, the vocabulary term `cov:confirmable` can be used.
+                A value of `true` indicates that a CON request SHOULD be used by the Consumer, while a value of `false` indicates that a NON request SHOULD be used instead (for potential usages of the vocabulary term see [[[#example-request-con-non]]]).
+                However, as the vocabulary term is just a hint, the consumer is not bound to this recommendation.
```
TallTed, line 221: Suggestions made against this content in `index.html` also apply to this content in `index.template.html`. But I don't understand why these files contain so much duplication.

danielpeintner: FYI: the duplication comes from the fact that based on `index.template.html` and _additional_ resources like ontologies etc. the _final_ `index.html` is generated. Hence, it is sufficient to fix `index.template.html` and tooling will do the rest.

TallTed: > Hence, it is sufficient to fix `index.template.html` and tooling will do the rest.

That would be true only if `index.html` contains nothing *in addition* to what is in `index.template.html`. Rather, "it is sufficient to fix `index.template.html` where it is duplicated in `index.html`, *and* fix `index.html` where it is not duplicated from `index.template.html`, and *then* tooling will do the rest".

Lacking anything highlighting the differences between `index.template.html` and `index.html`, reviewing these is thereby made a much larger effort.

danielpeintner: There is a [render](https://github.com/w3c/wot-binding-templates/blob/main/render.js) script in place that uses `index.template.html` and adds additional parts to it (based on other resources like ontologies). The result will be stored in `index.html`. Hence, changing `index.html` only does not have any effect, since a subsequent render task will override it again. See https://github.com/w3c/wot-binding-templates?tab=readme-ov-file#tooling

TallTed: So `index.template.heml` is not a *template*, and `index.html` is an _output artifact_, which means it shouldn't be under git management, unless it is stored to a different folder of the repo, which clearly indicates that its contents are such artifacts.

TallTed:
```suggestion
                In contrast to non-confirmable messages, confirmable messages are required to be acknowledged by the receiving endpoint and are otherwise retransmitted until an acknowledgement has been received.
                Using non-confirmable requests can be useful for sensor data (that gets sent repeatedly and therefore does not need to be transmitted reliably), while confirmable requests provide a certain reliability that is useful for retrieving values or invoking actions that should not fail.
            </p>
            <p>
                The vocabulary term `cov:confirmable` can be used to provide a hint whether CON or NON requests are expected to be better for accessing a CoAP resource associated with a form.
                A value of `true` encourages the Consumer to use CON requests, while a value of `false` encourages the Consumer to use NON requests. (For potential uses of the vocabulary term, see [[[#example-request-con-non]]].)
                Note that, as the vocabulary term is just a hint, the consumer is not bound to this recommendation.
```

egekorkan: > So `index.template.heml` is not a _template_, and `index.html` is an _output artifact_, which means it shouldn't be under git management, unless it is stored to a different folder of the repo, which clearly indicates that its contents are such artifacts.

We are slowly getting there as that was also what @lu-zero suggested :) Btw, this is the same in wot-thing-description repository. Sorry for the confusion and thank you for reviewing!

JKRhb: Thank you, @TallTed! I now applied your changes to the "template" document and re-rendered the resulting `index.html` file :)

bindings/protocols/coap/index.html
JKRhb: As always, these are very good points :) I was wondering whether the information regarding the regular sleep duration and the time frame during which the device is available would be enough to perform some kind of synchronization algorithm or whether more variables have to be given here – are you aware of any previous work that could be referenced here?

Also, your comments of course raise the question whether the addition of a `cov:confirmable` hint can be justified, or whether this should always be up to the consumer to decide. Looking at this from the Scripting API angle, I am not sure how/if the hint for the CoAP driver could be given in a protocol-agnostic way or whether the script has to be aware that CoAP will be used (in a specific context). Then again, we could also have certain semantic annotations that refer to the use case to help the consumer implementation choose the right message type, I suppose.

ektrah: In a nutshell, I would describe a TD as a list of instructions of the form: If you want to perform *&lt;this protocol-independent interaction&gt;*, then do *&lt;that protocol-specific interaction&gt;*.

A TD consumer should not need to know whether to use CoAP GET or POST, or CON or NON; it should just find the right form/instructions and follow them.

If the creator of a TD knows that (for example) CoAP GET with NON should always be used for a certain protocol-independent interaction, then I think there is nothing wrong with writing this into the TD. The TD consumer does not need to understand what GET with NON means; this is the job of the CoAP driver. But I think there needs to be an explanation/example/rule of thumb for the TD creator how to decide whether CON or NON should be specified in this way.

If the TD creator cannot make a blanket statement for an interaction, then I think the right way of doing things would be to include multiple sets of instructions (with CON or with NON or neither) and the TD consumer has to choose the right one in a protocol-independent way. This means there would have to be a protocol-independent vocabulary to describe possible intentions of the TD consumer, and the TD consumer chooses based on the current intention. This vocabulary could be, for example, additional (sub-)types of `op`, such as `readsensorproperty` or `invokefireandforgetaction`.

ektrah (approved):
@JKRhb I've done some "Verschlimmbesserungen". Please take a look 😺 

I think it would still be very good if there was an explanation/example/rule of thumb for the TD creator when to set `cov:confirmable` to `true` or `false`. This should not block merging, though.

JKRhb:
> @JKRhb I've done some "Verschlimmbesserungen". Please take a look 😺
> 

Thank you, @ektra, looks pretty good to me :)

> I think it would still be very good if there was an explanation/example/rule of thumb for the TD creator when to set `cov:confirmable` to `true` or `false`. This should not block merging, though.

Good point, I would try adding that in a follow-up PR if that is okay for everyone :)

egekorkan:
In the TD call, we agreed to merge this. However, @JKRhb will create a user story in the use cases repo for a more abstracted mechanism that works in multiple protocols

PR #394, opened 2025-02-03
chore: downgrade JSON Schema from 2019-09 to draft-07

fennibay:
wot-thing-description repo uses draft-07, as well; and 2019-09 is not well-supported in implementations. In TD 2.0 we should switch to 2020-12.

https://github.com/w3c/wot-thing-description/issues/1017

### DIFF
--- a/bindings/protocols/bacnet/bacnet.schema.json
+++ b/bindings/protocols/bacnet/bacnet.schema.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "https://json-schema.org/draft/2019-09/schema",
+    "$schema": "http://json-schema.org/draft-07/schema#",
     "title":"JSON Schema BACnet for TDs using BACnet Binding",
     "anyOf": [
         {
--- a/bindings/protocols/coap/coap.schema.json
+++ b/bindings/protocols/coap/coap.schema.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "https://json-schema.org/draft/2019-09/schema",
+    "$schema": "http://json-schema.org/draft-07/schema#",
     "anyOf": [
         {
             "$ref": "https://raw.githubusercontent.com/w3c/wot-thing-description/main/validation/td-json-schema-validation.json"
--- a/bindings/protocols/http/http.schema.json
+++ b/bindings/protocols/http/http.schema.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "https://json-schema.org/draft/2019-09/schema",
+    "$schema": "http://json-schema.org/draft-07/schema#",
     "anyOf": [
         {
             "$ref": "https://raw.githubusercontent.com/w3c/wot-thing-description/main/validation/td-json-schema-validation.json"
--- a/bindings/protocols/modbus/modbus.schema.json
+++ b/bindings/protocols/modbus/modbus.schema.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "https://json-schema.org/draft/2019-09/schema",
+    "$schema": "http://json-schema.org/draft-07/schema#",
     "anyOf": [
         {
             "$ref": "https://raw.githubusercontent.com/w3c/wot-thing-description/main/validation/td-json-schema-validation.json"
--- a/bindings/protocols/mqtt/mqtt.schema.json
+++ b/bindings/protocols/mqtt/mqtt.schema.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "https://json-schema.org/draft/2019-09/schema",
+    "$schema": "http://json-schema.org/draft-07/schema#",
     "anyOf": [
         {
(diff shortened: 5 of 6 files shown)


fennibay:
@egekorkan this handles the <https://github.com/w3c/wot-thing-description/issues/1017> for the short-term, please review.

danielpeintner (commented):
I think the changes look fine.

I just see lots of whitespace changes. Does this our tooling (linter) or the IDE you are using @fennibay ?

bindings/protocols/bacnet/bacnet.schema.json
```diff
@@ -1,5 +1,5 @@
 {
-    "$schema": "https://json-schema.org/draft/2019-09/schema",
+    "$schema": "http://json-schema.org/draft-07/schema#",
```
JKRhb, line 2: Hmm, I think we could still use https here (and in the other schema files), right?

danielpeintner: Mhh, it might have been an old AJV problem, but I would rather suggest going with http

see https://stackoverflow.com/questions/69133771/ajv-no-schema-with-key-or-ref-https-json-schema-org-draft-07-schema

fennibay:
> I think the changes look fine.
> 
> I just see lots of whitespace changes. Does this our tooling (linter) or the IDE you are using @fennibay ?

Probably my VS Code, sorry about that. I don't see this problem for other repos, but somehow always in this repo.

Should I rollback the whitespace changes? They are kind of nicer now 😄

bindings/protocols/bacnet/bacnet.schema.json
JKRhb: Yeah, just noticed that it also makes sense to use that URL since it is the same one as in the TD repository :) So we can consider this thread resolved :)

fennibay: I just followed what is in the thing-description repo: <https://github.com/w3c/wot-thing-description/blob/28353d79e50c0263f2924da9dfe40af77ce0cef5/validation/tm-json-schema-validation.json#L5>

We can do either way.

danielpeintner:
> > I think the changes look fine.
> > I just see lots of whitespace changes. Does this our tooling (linter) or the IDE you are using @fennibay ?
> 
> Probably my VS Code, sorry about that. I don't see this problem for other repos, but somehow always in this repo.
> 
> Should I rollback the whitespace changes? They are kind of nicer now 😄

I don't have a very strong opinion. Anyhow, I would rather like to see a linter rule in place to avoid changing back and forth every time someone else touches the file again ;-)
@egekorkan

JKRhb:
I think it is always nice to separate changes to the actual content from "stylistic" ones – so maybe the whitespace changes could be applied afterward as part of a PR that sets up formatter/linter rules?

ektrah:
It seems this repository does not have a `.gitattributes` file, which means the whitespace changes are probably actually line-endings changes.

JKRhb (approved):

JKRhb:
> Do we have a linter in the CI/CD pipeline to ensure that the schemas are actually valid according to the schema?

I think at the moment, there isn't any kind of validation being performed in the CI/CD pipeline, but that is definitely something we should add. Maybe we should open a tracking issue for, also listing the different aspects that should be covered.

danielpeintner:
> Do we have a linter in the CI/CD pipeline to ensure that the schemas are actually valid according to the schema?

You are raising a very good point and out of curiosity I copied the JSON schema files over to some online validators and found a problem. See https://github.com/w3c/wot-binding-templates/issues/397

egekorkan:
> It seems this repository does not have a `.gitattributes` file, which means the whitespace changes are probably actually line-endings changes.

We have the editorconfig file but that requires an editor with the plugin. I think we should be aligned across these kind of files and provide the format that works for the contributors.

relu91 (approved):

egekorkan:
I have checked https://json-schema.org/draft/2019-09/release-notes to see if there are 2019-09 features we have been using that will be lost/changed when we downgrade. Here are my findings:

- We have been using `definitions` even though we should have been using `$defs` as of 2019-09. By doing the wrong thing, we stay backwards compatible :D
- bacnet schema uses `format`, which is actually not supported in 2019-09. By going back, this will be enforced by validators. Thus, some bacnet TDs that were previously passing, can start failing. @fennibay is this ok? If it is ok, this PR can be merged and we should not have any implementation impact.

fennibay:
> * bacnet schema uses `format`, which is actually not supported in 2019-09. By going back, this will be enforced by validators. Thus, some bacnet TDs that were previously passing, can start failing. @fennibay is this ok? If it is ok, this PR can be merged and we should not have any implementation impact.

Yes this is OK, I was anyhow testing with validators that implement draft-07, and not 2019-09.

PR #395, opened 2025-02-03
Fix line endings

ektrah:


JKRhb (approved):

fennibay:
this is also fine for me, as an alternative please check #396, there I copied over .gitattributes from wot-thing-description, so that both are consistent.

fennibay (approved):

egekorkan:
@ektrah I do not understand why you removed the lf attribute? Also, going with #396 would make sense since it copies the settings in the TD repo. However, we did not have a long discussion about why we do it that way in the TD repo.

ektrah:
Different operating systems have different line ending conventions. Luckily, git has a perfect solution built in: by just having `* text=auto` in `.gitattributes`, on checkout all line endings will be converted to the current operating system's convention, and on commit all line endings will be normalized. I'm not sure why you wouldn't want to use this. 😺

egekorkan:
> Different operating systems have different line ending conventions. Luckily, git has a perfect solution built in: by just having `* text=auto` in `.gitattributes`, on checkout all line endings will be converted to the current operating system's convention, and on commit all line endings will be normalized. I'm not sure why you wouldn't want to use this. 😺

oh ok now I understand! That means we can simplify things a lot :) I will do a follow up PR in the TD repo as well

egekorkan (approved):

Issue #410, opened 2025-02-18
Add unique labels to registry mechanism items


egekorkan:
For now, we are using numbers per requirement/process within the registry. However, these numbers can shift over time. We should use some tags (labels etc.) to denote these sentences.

PR #411, opened 2025-02-18
Handling Conflicting Submissions

### LINKED ISSUE #398: Handling Binding Submissions for Existing Bindings
Raised at https://github.com/w3c/wot-binding-templates/issues/393#issuecomment-2622612507 and at https://github.com/w3c/wot-binding-templates/pull/378#discussion_r1938996685 , we can have the case where the a binding is submitted for a protocol/media type etc. that is already existing. We need to define how to handle this case. Some potential variations that come to mind:

1. The new submission improves the old one in an _objective_ way, e.g. adding a new vocabulary term. I think this is not a controversial one unless it comes from another entity.
2. A new submission comes when there is another one in the _initial_ state. I think this should be accepted since we can have two legitimate proposals for a binding as alternatives and based on implementation experience (or other criteria), one will be chosen.
3. A new submission comes when there is another one in the _stable_ state. If there is no visible improvement but just doing things differently, I would not be sure how to proceed. Maybe fine in _initial_ state and if there is enough implementation, we can review it. However, what happens when the old one also has enough implementations. What if the old one is from the SDO that owns that binding and vice versa?

From my point of view, only the 3rd point is difficult to solve in an objective way. We can require more reviewers and not take responsibility... Given that the URI scheme is a major way of differentiating bindings, we cannot let two bindings for the same URI scheme coexist with the same status.

EDIT: This is also part of https://github.com/w3c/wot-binding-templates/blob/a052686548bec6aed802e7e309fabe5232c1556e/registry-requirements.md#lifecycle under point 5 subitems. The opinions there are:

- @egekorkan At least per TD version, it should be avoided, but multiple initial entries should be allowed in case of diverging opinions to collect implementation feedback. If there is a drastically new binding of an existing one, two can be kept, but this should not be encouraged. This will cause issues with prefixes, URI scheme-based detection, other form elements, etc.
- @relu91  Additional fields in the form can indicate the binding version. Also, the prefix can be changed, e.g., `cov1.2` vs `cov1.3`
- @lu-zero Prefix versioning would require calling a new binding driver inside the implementation. Not using the prefix versioning implies JSON-LD processing, which not all implementations use.
- In the meeting, there was the idea to look at RFC 6838 but @egekorkan  could not find anything relevant in there.

egekorkan:
fixes #398

### DIFF
--- a/registry-requirements.md
+++ b/registry-requirements.md
@@ -71,7 +71,7 @@ DISCUSS: Clarify what happens when two ecosystems, OCF and LwM2M, use the CoAP b
 - Technical submission mechanism. How does a binding get submitted?
    - We work with issues only. The information for the entry format is submitted as a list. This way, non-W3C members can submit a binding. Reviews from the custodian happen on the issue. The submitter is expected to answer until the custodian makes a PR to add the binding to the registry or change its status.
    - A purpose built GitHub project for tracking submissions is used. When a submission comes in form an issue, it is automatically added to column "Binding Submitted". When the custodian and reviewers start looking at it, it goes to the "Under Review" column. If the review is in favour, the custodian makes the PR to add it to the registry and the issue goes to column "Accepted". If the review is not in favour, it goes to the column "Rejected". All these changes are reflected as comments in the original issue.
-- Technical Lifecycle Change Mechanism: 
+   - In case of potential conflict with another entry, the reviewer MUST mark the new submission accordingly. As two bindings that do the same are not allowed, either the old one MUST be deprecated or the new one MUST be rejected. See also point 13 under "Requirements on the Submitted Document".
 - Status values: Initial -> Current -> Superseded or Obsolete
   - Definitions:
     - Initial: Document is correctly written but no implementation experience has been necessarily documented.
@@ -86,8 +86,6 @@ DISCUSS: Clarify what happens when two ecosystems, OCF and LwM2M, use the CoAP b
 - What are the requirements for transitioning from one value to another? See the "Requirements on the Submitted Document" section as well.
 - Versioning of registry entries (see https://github.com/w3c/wot/tree/main/registry-analysis#versioning)
   - Ege: We do not allow updates to a registry document's content. A new version of a binding is a resubmission and optional deprecation of the old one. However, new features need new implementations, so it is not a completely new registration.
-  - DISCUSS: Do we allow two versions of a binding to stay in the registry?
-    - See https://github.com/w3c/wot-binding-templates/issues/398
 - Deletion and deprecation (see https://github.com/w3c/wot/tree/main/registry-analysis#deletion-and-deprecation-of-registry-entries)
   - No entry is ever deleted. Deprecated entries are moved to another table or are clearly marked deprecated, colored differently, and moved to the bottom.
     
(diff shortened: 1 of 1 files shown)


registry-requirements.md
```diff
@@ -71,7 +71,7 @@ DISCUSS: Clarify what happens when two ecosystems, OCF and LwM2M, use the CoAP b
...
    - We work with issues only. The information for the entry format is submitted as a list. This way, non-W3C members can submit a binding. Reviews from the custodian happen on the issue. The submitter is expected to answer until the custodian makes a PR to add the binding to the registry or change its status.
    - A purpose built GitHub project for tracking submissions is used. When a submission comes in form an issue, it is automatically added to column "Binding Submitted". When the custodian and reviewers start looking at it, it goes to the "Under Review" column. If the review is in favour, the custodian makes the PR to add it to the registry and the issue goes to column "Accepted". If the review is not in favour, it goes to the column "Rejected". All these changes are reflected as comments in the original issue.
-- Technical Lifecycle Change Mechanism: 
+   - In case of potential conflict with another entry, the reviewer MUST mark the new submission accordingly. As two bindings that do the same are not allowed, either the old one MUST be deprecated or the new one MUST be rejected. See also point 13 under "Requirements on the Submitted Document".
```
danielpeintner, line 74:
```suggestion
   - In case of potential conflict with another entry, the reviewer MUST mark the new submission accordingly. As two bindings that do the same are not allowed, either the old one MUST be deprecated or the new one MUST be rejected. See also aspects under "Requirements on the Submitted Document".
```

I think it is very likely the number will go out of sync

egekorkan: Yes and that is why there will be a follow up PR to address #410

TallTed, line 74: Is this about "potential" conflict? or *actual* conflict? I think the latter, so 
```suggestion
   - If a new entry conflicts with another entry, the reviewer MUST mark the new submission accordingly. As two bindings that do the same are not allowed, either the old one MUST be deprecated or the new one MUST be rejected. See also point 13 under "Requirements on the Submitted Document".
```

registry-requirements.md
```diff
@@ -125,5 +123,5 @@ What does the binding have to contain to go into the table
   - See https://github.com/w3c/wot-binding-templates/issues/404
 12. DISCUSS: Where should discussions on implementation experience be collected?
   - See https://github.com/w3c/wot-binding-templates/issues/403
-13. DISCUSS: How to handle overlaps between two bindings in the review process?
-  - See https://github.com/w3c/wot-binding-templates/issues/398 and https://github.com/w3c/wot-binding-templates/issues/401
+13. The binding entry SHOULD NOT conflict with other entries in the registry by redefining the same concepts such as redefining the URI Scheme, the vocabulary terms or the default assignments. If the previously stable binding is being improved upon by the same organization, that previous binding MUST be deprecated once the new one reaches the **stable** status.
```
TallTed, line 126:
```suggestion
13. The binding entry SHOULD NOT conflict with other entries in the registry by redefining concepts, such as the URI Scheme, the vocabulary terms, or the default assignments. If a previously stable binding is being improved upon by the same organization, that previous binding MUST be deprecated once the new one reaches the **stable** status.
```

PR #420, opened 2025-03-11
Binding Dependencies

### LINKED ISSUE #401: Handling Bindings that depend on other bindings
As part of https://github.com/w3c/wot-binding-templates/pull/378 at https://github.com/w3c/wot-binding-templates/blob/a052686548bec6aed802e7e309fabe5232c1556e/registry-requirements.md#entry-format under the paragraph of point 8, we have started a discussion on the cases where a binding (e.g. LWM2M) depends on the CoAP binding. How should this dependency be handled and how should it be highlighted in the registry?

egekorkan:
fixes #401 

By the way, the correct terminology is simply dependencies and dependents. Not sure why something so obvious didn't come to mind :D


egekorkan:
Call of 12 March: Agreed to merge

PR #428, opened 2025-06-16
fix: model BACnet relinquish explicitly

fennibay:
So far we hadn't modelled the relinquish operation explicitly, which is problematic as it requires the WoT consumer to write null to a PropertyAffordance to clear a priority, even though its DataSchema doesn't allow it.

Options I've evaluated:

1. Leave AS IS -> implicit knowledge of the WoT Consumer is needed
2. Make the PropertyAffordance schema as { oneOf: null or number } ->This would imply that the value could be null when reading or observing, which is wrong.
3. Add another URI variable relinquish: boolean. -> This PR's solution, even though it still requires a dummy write value to be provided, which will be ignored
4. Using ActionAffordance -> creates unnecessary dual entities of the same BACnet Object, and the WoT Consumers needs to deal with the complexity of handling that they deal with the same entity. Additionally it doesn't gracefully handle WoT Consumers that call writeproperty without a priority.

@egekorkan @ektrah @mjkoster please have a look.

### DIFF
--- a/bindings/protocols/bacnet/bacnet-model.ttl
+++ b/bindings/protocols/bacnet/bacnet-model.ttl
@@ -43,7 +43,7 @@ Wherever possible, this model refers to ANSI/ASHRAE Standard 135-2020 for BACnet
 
 hctl:hasTarget rdfs:comment """
 Should be of form bacnet://<device>/<objecttypenr>,<objectinsttr>[/<property>[/<index>]]
-Additional parameters can be specified via URI variable syntax, e.g. ?commandPriority=13 for command priority, ?relinquish for relinquishing a priority or ?covIncrement=0.5 for COV increment
+Additional parameters can be specified via URI variable syntax, e.g., `?commandPriority=13` to change command priority, `?relinquish` to relinquish a priority, or `?covIncrement=0.5` to increment COV.
 """@en ;
                rdfs:seeAlso "ASHRAE 135-2020 BACnet Q.8 BACnet URI Scheme" .
 
@@ -103,7 +103,7 @@ bacv:CommandPriority rdf:type owl:Class ;
                            # Default priority for writing is 16, ref. ASHRAE 135-2020 BACnet 15.9.1.1.5 Priority
                            owl:hasValue 16
                        ] ;
-                       rdfs:comment "Specifies the writing priority for BACnet objects with commanding, i.e. priority array."@en ;
+                       rdfs:comment "Specifies the writing priority for BACnet objects with commanding, i.e., priority array."@en ;
                        rdfs:seeAlso "ASHRAE 135-2020 BACnet 19.2 Command Prioritization" .
 
 bacv:Relinquish rdf:type owl:Class ;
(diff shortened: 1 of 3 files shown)


bindings/protocols/bacnet/index.html
```diff
@@ -1439,7 +1465,7 @@ <h2 id="examples">Examples</h2>
                             },
                             {
                                 "op": [ "writeproperty" ],
-                                "href": "bacnet://5/2,1?commandPriority={writePriority}",
+                                "href": "bacnet://5/2,1?commandPriority={writePriority}{&relinquish}",
```
egekorkan, line 1468: `&` should be outside of the `{}` no?

egekorkan (commented):
I thought I understood this from the discussion but then when I looked into the PR, I got confused. @fennibay could you give a real-life example?

bindings/protocols/bacnet/index.html
danielpeintner: I think it is even more different. I am not an [rfc6570](https://www.rfc-editor.org/rfc/rfc6570) expert  but playing around with https://www.utilities-online.info/uritemplate seems to work fine with

[`bacnet://5/2,1?commandPriority={writePriority}&relinquish={relinquishValue}`](bacnet://5/2,1?commandPriority={writePriority}&relinquish={relinquishValue})

and as input 

```
{
  "writePriority": 16,
  "relinquishValue": true
}
```

creates

`bacnet://5/2,1?commandPriority=16&relinquish=true`

egekorkan:
Until @mjkoster gives ok (or next TD call), we cannot merge the PR as it is only provided and approved by the same member organization.

bindings/protocols/bacnet/bacnet-model.ttl
```diff
@@ -43,7 +43,7 @@ Wherever possible, this model refers to ANSI/ASHRAE Standard 135-2020 for BACnet
 
 hctl:hasTarget rdfs:comment """
 Should be of form bacnet://<device>/<objecttypenr>,<objectinsttr>[/<property>[/<index>]]
-Additional parameters can be specified via URI variable syntax, e.g. ?commandPriority=13 for command priority or ?covIncrement=0.5 for COV increment
+Additional parameters can be specified via URI variable syntax, e.g. ?commandPriority=13 for command priority, ?relinquish for relinquishing a priority or ?covIncrement=0.5 for COV increment
```
TallTed, line 46: I know this isn't markdown, but the backticks provide additional clarity even when they are displayed and not transformed into `<code>` or similar markup.
```suggestion
Additional parameters can be specified via URI variable syntax, e.g., `?commandPriority=13` to change command priority, `?relinquish` to relinquish a priority, or `?covIncrement=0.5` to increment COV.
```

egekorkan:
TD Call of Today:
- We have reviewed it and the TF is ok with it.
- There are some small bug fixes and clarifications in addition to the adding the variable to explicitly describe the relinquish feature of BACnet (which already exists in BACnet systems)

PR #431, opened 2025-07-28
fix: update profinet schema to correct references and add new types

alicanipek:
This fix updates the bindings/protocols/profinet/profinet.schema.json file.
- fixes profv:enumeratedValue
- adds object as profv:type


alicanipek:
Thank you all for your comments.
@relu91 I reverted json schema change.

PR #432, opened 2025-07-30
Adapting to Registry and Binding Word Usage

egekorkan:
After the decision to not use the word "binding template" but just binding and with the upcoming registry, I am changing the title, links and a bunch of other small mentions of the words.

### DIFF
--- a/bindings/index.template.html
+++ b/bindings/index.template.html
@@ -40,7 +40,7 @@
                   data: [
                       {
                       value: "BINDINGNAME JSON Schema",
-                      href: "bindingname.schema.json"
+                      href: "template.schema.json"
                       }
                   ]
               }
--- a/bindings/payloads/xml/index.html
+++ b/bindings/payloads/xml/index.html
@@ -41,15 +41,6 @@
                         href: " https://w3c.github.io/wot-thing-description/#bindings"
                     }
                 ]
-            },
-            {
-                key: "JSON Schema",
-                data: [
-                    {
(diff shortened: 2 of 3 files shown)


bindings/index.template.html
egekorkan, line 1: This document should be reviewed. The others are just copying the changes in here

egekorkan:
The PR is ready for reviews. @JKRhb @mjkoster @relu91 @fennibay @ektrah @Kaz040 please confirm that the changes are fine with you. It is basically changing "Binding Template" to "Binding" in the relevant places.

bindings/index.template.html
```diff
@@ -16,51 +16,70 @@
...
+                        }
+                    ]
+                },
                 {
                   key: "JSON Schema",
                   data: [
                       {
                       value: "BINDINGNAME JSON Schema",
-                      href: "template.schema.json"
+                      href: "bindingname.schema.json"
```
mjkoster, line 43: Isn't this the template file for the JSON Schema that goes in a binding? Do we need to change the name of this file if it doesn't refer to a "binding template"?

egekorkan: Good catch... It should be reverted and the other references should reference https://github.com/w3c/wot-binding-templates/blob/main/bindings/protocols/template.schema.json instead

egekorkan: Done

egekorkan:
TD Call 21 Aug: The TF agrees to merge this after a review.

PR #451, opened 2025-11-14
docs(BACnet): remark why data link layers are left out

fennibay:
This is already our design, so no change in content. It is helpful for readers looking for data link layer specifics in this binding.

### DIFF
--- a/bindings/protocols/bacnet/index.html
+++ b/bindings/protocols/bacnet/index.html
@@ -161,9 +161,9 @@
       <p>
         This binding leaves out BACnet data link layer details (e.g. IP, MS/TP, Secure Connect), as the BACnet device
         identifier provides sufficient addressing in a properly configured network. This approach avoids complexities
-        that arise such as when WoT consumer and WoT thing use different data links connected via BACnet routers.
-        This design philosophy mirrors the HTTP protocol binding, which focuses on HTTP semantics
-        rather than underlying IP infrastructure such as routing and name resolution.
+        that arise such as when WoT consumer and WoT thing use different data links connected via BACnet routers. This
+        design philosophy mirrors the HTTP protocol binding, which focuses on HTTP semantics rather than underlying IP
+        infrastructure such as routing and name resolution.
       </p>
       <p>
         WoT Interaction Affordances are mapped to BACnet Properties. If the BACnet property-identifier is not present in
--- a/bindings/protocols/bacnet/index.html
+++ b/bindings/protocols/bacnet/index.html
@@ -161,9 +161,9 @@
       <p>
         This binding leaves out BACnet data link layer details (e.g. IP, MS/TP, Secure Connect), as the BACnet device
         identifier provides sufficient addressing in a properly configured network. This approach avoids complexities
-        that arise such as when WoT consumer and WoT thing use different data links connected via BACnet routers. This
+        that arise, such as when WoT consumer and WoT thing use different data links connected via BACnet routers. This
         design philosophy mirrors the HTTP protocol binding, which focuses on HTTP semantics rather than underlying IP
-        infrastructure such as routing and name resolution.
+        infrastructure, such as routing and name resolution.
       </p>
       <p>
         WoT Interaction Affordances are mapped to BACnet Properties. If the BACnet property-identifier is not present in
--- a/bindings/protocols/bacnet/index.html
+++ b/bindings/protocols/bacnet/index.html
@@ -159,7 +159,7 @@
         onboarding, or device management.
       </p>
       <p>
-        This binding leaves out BACnet data link layer details (e.g. IP, MS/TP, Secure Connect), as the BACnet device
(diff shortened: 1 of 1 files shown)


bindings/protocols/bacnet/index.html
```diff
@@ -158,6 +158,13 @@ <h2>Binding Scope</h2>
         affordances exposed by BACnet devices. It is not intended that this binding be used for device commissioning,
         onboarding, or device management.
       </p>
+      <p>
+        This binding leaves out BACnet data link layer details (e.g. IP, MS/TP, Secure Connect), as the BACnet device
```
TallTed:
```suggestion
        This binding leaves out BACnet data link layer details (e.g., IP, MS/TP, Secure Connect), as the BACnet device
```

egekorkan (approved):

bindings/protocols/bacnet/index.html
```diff
@@ -158,6 +158,13 @@ <h2>Binding Scope</h2>
         affordances exposed by BACnet devices. It is not intended that this binding be used for device commissioning,
         onboarding, or device management.
       </p>
+      <p>
+        This binding leaves out BACnet data link layer details (e.g., IP, MS/TP, Secure Connect), as the BACnet device
+        identifier provides sufficient addressing in a properly configured network. This approach avoids complexities
+        that arise, such as when WoT consumer and WoT thing use different data links connected via BACnet routers. This
+        design philosophy mirrors the HTTP protocol binding, which focuses on HTTP semantics rather than underlying IP
+        infrastructure, such as routing and name resolution.
```
benfrancis, line 166: If I have understood this correctly I don't think this is really comparable, and IMO all non-IP protocols should be considered out of scope for all WoT bindings.

The mission of the Web of Things Working Group is to "counter the fragmentation of the Internet of Things (IoT)" which is already a big enough scope, without extending this to non-Internet protocols. This is why we agreed early on not to create bindings for other non-IP protocols like Zigbee and Z-Wave.

I think that you can argue that a device using BACnet/IP could be considered an IoT device, but as I understand it a BACNet device which uses MS/TP (RS-485) as a transport is not really an IoT device unless a bridge or gateway bridges it to an IP-based protocol - in which case it is the affordances exposed by the gateway which should be described by a WoT TD.

I would therefore suggest that the binding should be limited to devices which use the Internet Protocol as a transport, because otherwise both IP-based and non-IP based devices could be described in a TD using the `bacnet://` URI scheme and a WoT Consumer has no way of distinguishing between IP-based devices it can communicate with and non-IP devices it can't communicate with.

If I've misunderstood the intention of this paragraph (or how BACnet works) then please do let me know.

egekorkan: One general thing: at Siemens, we do describe non-IP devices with TDs and it seems to work. I had the same opinion as you before but now I don't see a big problem. The Consumer can understand if they can talk with the device and if it doesn't support the protocol, it can leave it out. Not supporting can be due to missing hw interface or not implementing the driver, same result.

So from my understanding for BACnet, you don't necessarily know the transport unless you query some other properties. They are addressed with the same object id no matter the transport.

benfrancis: @egekorkan wrote:
> at Siemens, we do describe non-IP devices with TDs and it seems to work.

I don't doubt that it works for your specific use cases, but I think this is dangerous scope creep.

egekorkan: > I don't doubt that it works for your specific use cases, but I think this is dangerous scope creep.

With the abstraction we have at the moment with bindings, which can be developed by whomever, we are not losing the scope of the WG work. I would not prioritize those protocols with all the other work we have but if somebody wants to do it, it is fine. Here are the non-IP protocols I have seen in TDs (not just Siemens btw) that I don't need to care about since they just work with the abstraction we have:

- Lorawan
- Bluetooth (Gannt)
- KNX (not KNX IoT)
- Modbus RTU
- Apogee P2 (might be IP layer in some cases?)
- Mbus

danielpeintner: Netconf ? uses RPC calls as far as I know...

benfrancis: OK, I think we will have to agree to disagree on this point because it boils down to our different use cases of what I would describe as the Web _of_ Things vs. the Web _about_ Things (i.e. whether Things are part of the Web or just described by it). Hopefully both approaches can co-exist without the complexity of Thing Descriptions exploding, and time will tell which one leads to the most cross-vendor interoperability.

fennibay: @benfrancis thanks for sharing your concerns.

I'm fully with you that integrating non-IP protocols shall not complicate web of things. In the worst-case scenario it should be constrained to the specific protocol binding. In the BACnet case, we chose IMO the simplest option which doesn't bring the data link layer specifics to the binding at all.

PR #472, opened 2026-08-19
Modbus Address Range Explanation

### LINKED ISSUE #468: Modbus address range
In the explanation of the functions, there is text like:

- Usually in the address range 00001-09999
- Address range 10001-19999

I am not sure we should mention this in the spec as it is creating confusion. If one sees `href:40003` does it mean they should sent to register 3? I think it is better to just use 3 in the href. Also, putting `usually` or not is creating a false interpretation about default behavior. @relu91 I remember discussing this but could not find an issue. What is your point of view?

egekorkan:
fixes #468

### DIFF
--- a/bindings/protocols/modbus/index.html
+++ b/bindings/protocols/modbus/index.html
@@ -186,7 +186,7 @@
 
             <tr>
               <td><code>modv:address</code></td>
-              <td>Specifies the starting address of the Modbus operations</td>
+              <td>Specifies the starting address of the Modbus operations. Between 0 and 65535</td>
               <td>required</td>
               <td><a href="http://www.w3.org/2001/XMLSchema#integer">integer</a></td>
             </tr>
--- a/bindings/protocols/modbus/index.html
+++ b/bindings/protocols/modbus/index.html
@@ -196,8 +196,9 @@
             <tr>
               <td><code>modv:unitID</code></td>
               <td>
-                The Unit ID is technically not needed for ModbusTCP, since the IP-address works as unique identifier,
-                but for compatibility reasons it MUST be included
+                Sometimes also called the Slave or Server ID, the Unit ID is technically not needed for ModbusTCP, since
+                the IP-address works as unique identifier, but for compatibility reasons it MUST be included. Mostly
+                between 1 and 247 but 255 is technically possible
               </td>
               <td>required</td>
               <td><a href="http://www.w3.org/2001/XMLSchema#integer">integer</a></td>
@@ -246,8 +247,8 @@
             <tr>
               <td><code>modv:timeout</code></td>
               <td>
-                Modbus response maximum waiting time. Defines how much time the runtime should wait until it receives a
-                reply from the device.
+                Maximum waiting time for a response. Defines how much time the runtime should wait until it receives a
+                reply from the device. Defined as integer of milliseconds.
               </td>
               <td>optional</td>
               <td><a href="http://www.w3.org/2001/XMLSchema#integer">integer</a></td>
--- a/bindings/protocols/modbus/index.html
+++ b/bindings/protocols/modbus/index.html
@@ -431,7 +431,7 @@
               <td><code>readDiscreteInput</code></td>
               <td>Read Discrete Inputs</td>
               <td>2</td>
-              <td>Read Physical Discrete Inputs (bit access). Address range 10001-19999</td>
+              <td>Read Physical Discrete Inputs (bit access). Usually in the address range 10001-19999</td>
             </tr>
 
             <tr></tr>
@@ -449,7 +449,7 @@
               <td><code>readHoldingRegisters</code></td>
               <td>Read Multiple Holding Registers</td>
               <td>3</td>
-              <td>Read Multiple Holding Registers (16 bit access). Address range 40001-49999</td>
+              <td>Read Multiple Holding Registers (16 bit access). Usually in the address range 40001-49999</td>
             </tr>
 
             <tr></tr>
@@ -458,7 +458,7 @@
               <td><code>readInputRegisters</code></td>
               <td>Read Multiple Input Registers</td>
               <td>4</td>
-              <td>Read Multiple Physical Input Registers (16 bits). Address range 30001-39999</td>
+              <td>Read Multiple Physical Input Registers (16 bits). Usually in the address range 30001-39999</td>
             </tr>
 
             <tr></tr>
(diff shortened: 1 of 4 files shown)


bindings/protocols/modbus/ontology.ttl
```diff
@@ -144,7 +144,7 @@ binding:assignment rdf:type owl:AnnotationProperty .
 :hasAddress rdf:type owl:DatatypeProperty ;
             rdfs:domain :Request ;
             rdfs:range xsd:integer ;
-            rdfs:comment "Specifies the starting address of the Modbus operations"@en ;
+            rdfs:comment "Specifies the starting address of the Modbus operations. Between 0 and 65535"@en ;
```
danielpeintner: Wouldn't that be `xsd:unsignedShort`
see https://www.w3.org/TR/xmlschema-2/#unsignedShort

egekorkan: Yes this is correct! Good catch. I am looking at other terms too.

bindings/protocols/modbus/index.html
```diff
@@ -494,12 +497,19 @@ <h3>Function</h3>
...
 
             <tr></tr>
           </tbody>
         </table>
+        <p class="note">
+          The address ranges indicated are based on the observations of the Modbus community and are not part of the
+          official Modbus specification. While you can use any address within the 0-65535 range, it is recommended for
+          new devices to adhere to these ranges for better compatibility with existing Modbus devices and systems. A WoT
+          Consumer application should not assume that a specific address range is used for a particular type of register
+          or coil, and should always use the Thing Description as the source of truth.
```
egekorkan: We should add a [warning](https://respec.org/docs/#warning) that there are libraries that assume that a register with address 3 is actually at address 40003. However, TD document the device as is and if you are using such a library/SDK, you need to post process the TD before handing it over to the library.

bindings/protocols/modbus/index.html
```diff
@@ -494,12 +497,19 @@ <h3>Function</h3>
               <td><code>writeSingleHoldingRegister</code></td>
               <td>Write Single Holding Register</td>
               <td>6</td>
-              <td>Write Single HoldingRegister (internal bits). Address range 40001-49999</td>
+              <td>Write Single HoldingRegister (internal bits). Usually in the address range 40001-49999</td>
```
egekorkan: Instead of "Usually", "Conventionally" should be used

egekorkan:
Both feedbacks are incorporated now. The warning box is warning me a bit too much but I am ok with it.

bindings/protocols/modbus/index.html
```diff
@@ -412,7 +412,9 @@ <h3>Function</h3>
               <td><code>readCoil</code></td>
               <td>Read Single Coil</td>
               <td>1</td>
-              <td>Read a single coil (i.e. boolean/bit access) value. Usually in the address range 00001-09999</td>
+              <td>
+                Read a single coil (i.e. boolean/bit access) value. Conventionally in the address range 00001-09999
```
TallTed, line 416:
```suggestion
                Read a single coil (i.e., boolean/bit access) value. Conventionally in the address range 00001-09999
```

bindings/protocols/modbus/index.html
```diff
@@ -497,18 +500,20 @@ <h3>Function</h3>
...
-        <p class="note">
+        <p class="warning">
           The address ranges indicated are based on the observations of the Modbus community and are not part of the
           official Modbus specification. While you can use any address within the 0-65535 range, it is recommended for
           new devices to adhere to these ranges for better compatibility with existing Modbus devices and systems. A WoT
           Consumer application should not assume that a specific address range is used for a particular type of register
-          or coil, and should always use the Thing Description as the source of truth.
+          or coil, and should always use the Thing Description as the source of truth. However, pay attention that your
+          Modbus library does not have any expectation and thus limitations on the address range. In those cases, you
+          should post process the TD before handing it over to the Modbus library.
```
TallTed, line 516:
```suggestion
          Modbus library does not have any expectations and thus limitations on the address range. In those cases, you
          should post-process the TD before handing it over to the Modbus library.
```

bindings/protocols/modbus/index.template.html
```diff
@@ -261,12 +261,14 @@ <h3>Function</h3>
...
-        <p class="note">
+        <p class="warning">
           The address ranges indicated are based on the observations of the Modbus community and are not part of the
           official Modbus specification. While you can use any address within the 0-65535 range, it is recommended for
           new devices to adhere to these ranges for better compatibility with existing Modbus devices and systems. A WoT
           Consumer application should not assume that a specific address range is used for a particular type of register
-          or coil, and should always use the Thing Description as the source of truth.
+          or coil, and should always use the Thing Description as the source of truth. However, pay attention that your
+          Modbus library does not have any expectation and thus limitations on the address range. In those cases, you
+          should post process the TD before handing it over to the Modbus library.
```
TallTed, line 271:
```suggestion
          Modbus library does not have any expectations and thus limitations on the address range. In those cases, you
          should post-process the TD before handing it over to the Modbus library.
```

bindings/protocols/modbus/ontology.html
```diff
@@ -715,7 +715,9 @@ <h4>Input Register</h4>
         <section>
           <h4>Read Single Coil</h4>
           <p>IRI: <code>https://www.w3.org/2019/wot/modbus#readCoil</code></p>
-          <span>Read a single coil (i.e. boolean/bit access) value. Usually in the address range 00001-09999</span>
+          <span
+            >Read a single coil (i.e. boolean/bit access) value. Conventionally in the address range 00001-09999</span
```
TallTed, line 719:
```suggestion
            >Read a single coil (i.e., boolean/bit access) value. Conventionally in the address range 00001-09999</span
```

bindings/protocols/modbus/ontology.ttl
```diff
@@ -478,7 +478,7 @@ The data request field may be nonexistent (zero length) in certain kind of reque
                    :Function ;
           :readFrom :Coil ;
           :hasCodeNumber 1 ;
-          rdfs:comment "Read a single coil (i.e. boolean/bit access) value. Usually in the address range 00001-09999"@en ;
+          rdfs:comment "Read a single coil (i.e. boolean/bit access) value. Conventionally in the address range 00001-09999"@en ;
```
TallTed, line 481:
```suggestion
          rdfs:comment "Read a single coil (i.e., boolean/bit access) value. Conventionally in the address range 00001-09999"@en ;
```

PR #476, opened 2026-08-20
TD Schema Reference Fix

egekorkan:
All the schemas were referring to the main branch schema in the TD repo. That is subject to breaking changes. I am fixing to the stable link.

Additionally, one modbus example was using `pollingRate` instead of `pollingTime` and the context URI in modbus was the github.io URL. While that is correct, it is not used anywhere else and would cause issues somewhere in the end.

### DIFF
--- a/bindings/protocols/bacnet/bacnet.schema.json
+++ b/bindings/protocols/bacnet/bacnet.schema.json
@@ -3,7 +3,7 @@
   "title": "JSON Schema BACnet for TDs using BACnet Binding",
   "anyOf": [
     {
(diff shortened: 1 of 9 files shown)


ektrah (approved):

JKRhb (approved):

fennibay (approved):

danielpeintner:
I think the fix itself is fine. Anyhow, a related fix should go into the JSON schema as well.. I believe

Looking at https://www.w3.org/2022/wot/td-schema/v1.1 we see the `$id` might need to be updated as well.

grafik

The same goes for the TM JSON schema at https://www.w3.org/2022/wot/tm-schema/v1.1

egekorkan:
The resources are also underway: https://github.com/w3c/wot-resources/pull/30

PR #481, opened 2026-08-31
Binding registry config in Respec

egekorkan:
simple update to all bindings' link to registry. Also in profinet, there were two `localBiblio` objects so that should reduce respec warnings.


ektrah (approved):

danielpeintner (approved):
LGTM

Kaz040 (approved):

PR #484, opened 2026-09-14
Draft EtherNet/IP protocol binding

barnstee:


egekorkan:
- We need to check the potentially reserved characters in the URI Scheme
- In example 3, we need to check if `[` syntax is correct.
- Example class of the uri scheme definition should be fixed
- `eip` prefix should be changed so that it is not mixed up with the URI scheme (recommendation)
- @danielpeintner to review the xsd datatypes as some types don't seem to exist in XML Schema. https://www.w3.org/TR/xmlschema11-2/#built-in-datatypes can be used for reference.
- Section 5.3 can be removed since it is part of the TD core spec
- Security can be revisited based on experience with other controllers. So far no evidence. Maybe newer versions or extended versions have security.
- example 9, query param can be removed. remnant of example from above

egekorkan:
We had a look during the TD call and all the feedback is addressed. @egekorkan will have a look offline.

